PR #1, opened 2018-06-08
Close netlink connection after flushing commands

mdlayher:
Not sure if GC would take care of this or not, but it can't hurt to clean up.


stapelberg:
Thanks!

PR #6, opened 2019-01-04
Implement DelTable

twitchy-jsonp:
Using my googley github account cuz CLABot.

Background: Am working on a namespace-based thingy to sandbox processes. I'm creating a veth for each 'box' and punting one end into the processes' network namespace. I'm using nftables to NAT the traffic, so I create a table for each box.

When the process completes and is torn down, its best to cleanup the rules used for the now-deleted IP addresses, hence the implementation of `DelTable`.

I determined the netlink packet structure by looking at `libnftnl` source (specifically `nft_ruleset_table_build_msg`, which converts a 'flush' command into `NFT_MSG_DEL*` commands based on the information provided. Given how simple it is I havent put in any tests.

### DIFF
--- a/nftables.go
+++ b/nftables.go
@@ -329,6 +329,21 @@ func (cc *Conn) FlushRuleset() {
 	})
 }
 
+// DelTable deletes a specific table, along with all chains/rules it contains.
+func (cc *Conn) DelTable(t *Table) {
+	data := cc.marshalAttr([]netlink.Attribute{
+		{Type: unix.NFTA_TABLE_NAME, Data: []byte(t.Name + "\x00")},
+		{Type: unix.NFTA_TABLE_FLAGS, Data: []byte{0, 0, 0, 0}},
+	})
+	cc.messages = append(cc.messages, netlink.Message{
+		Header: netlink.Header{
+			Type:  netlink.HeaderType((unix.NFNL_SUBSYS_NFTABLES << 8) | unix.NFT_MSG_DELTABLE),
+			Flags: netlink.HeaderFlagsRequest | netlink.HeaderFlagsAcknowledge,
+		},
+		Data: append(extraHeader(uint8(t.Family), 0), data...),
+	})
+}
+
 // AddTable adds the specified Table. See also
 // https://wiki.nftables.org/wiki-nftables/index.php/Configuring_tables
 func (cc *Conn) AddTable(t *Table) *Table {


nftables.go
```diff
@@ -329,6 +329,21 @@ func (cc *Conn) FlushRuleset() {
...
+// DelTable deletes a specific table, aling with all chains/rules it contains.
+func (cc *Conn) DelTable(t *Table) {
+	data := cc.marshalAttr([]netlink.Attribute{
+		{Type: unix.NFTA_TABLE_NAME, Data: []byte(t.Name + "\x00")},
+		{Type: unix.NFTA_TABLE_FLAGS, Data: []byte{0, 0, 0, 0}},
+	})
+	cc.messages = append(cc.messages, netlink.Message{
+		Header: netlink.Header{
+			Type:  netlink.HeaderType((unix.NFNL_SUBSYS_NFTABLES << 8) | unix.NFT_MSG_DELTABLE),
+			Flags: netlink.HeaderFlagsRequest | netlink.HeaderFlagsAcknowledge | netlink.HeaderFlagsCreate,
```
stapelberg: Are you sure about the flags? In libnftnl/examples/nft-table-del.c, I only see HeaderFlagsAcknowledge.

Maybe a test would be worthwhile after all :)

twitchyliquid64: I'm pretty certain `NLM_F_REQUEST` is set at a higher level in the library - FWICT the kernel only handles 'request' messages and [rejects all others](https://github.com/torvalds/linux/blob/9a76aba02a37718242d7cdc294f0a3901928aa57/net/netfilter/nfnetlink.c#L378).

Likewise `NLM_F_ACK` indicates that the kernel should respond with error messages (as opposed to just setting BATCH_FAILURE status. We definitely want this behavior for in the future - it would be nice for `Flush()` to properly detect errors and return a useful error structure.

twitchyliquid64: I think `HeaderFlagsCreate` is superfluous thou.

stapelberg: Thanks for the details. Can you check if it works without HeaderFlagsCreate and update the PR if so please? :)

twitchyliquid64:
Manually tested without the `HeaderFlagsCreate` on a 4.20.0 kernel. Should work on others too (I dont think the NETLINK infra has changed much since nftables went in).

Should be good to go, unless you insist on tests :)

twitchyliquid64:
This is automatic on github; just press the 'Merge and squash commit' green button thingy.

stapelberg:
That will retain the commit messages, though, and I’d prefer messages such as “I cant spell” to not end up in the history :)

twitchyliquid64:
Oh lol whelp!! I'll do a hard squash in the morning then. XD

twitchy-jsonp:
Okies G2G

stapelberg:
Thanks!

PR #7, opened 2019-01-14
Implement verdict expression

twitchy-jsonp:
Soooooooo I was trying to do some firewalling now that iptables silently rekts itself if you use it on the same host as you are using nftables. To that end, I have implemented verdicts.

I've been quite liberal in leaving in comments about where to find information (particularly references to other code, and describing the layout of the structure). This is because there is literally no documentation online about this (it took a while to get this working); I feel it needs to be somewhere (but it also looks a little messy; if you have suggestions on the layout, let me know).

### DIFF
--- a/expr/verdict.go
+++ b/expr/verdict.go
@@ -1,0 +1,89 @@
+// Copyright 2018 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package expr
+
+import (
+	"fmt"
+
+	"github.com/google/nftables/binaryutil"
+	"github.com/mdlayher/netlink"
+	"golang.org/x/sys/unix"
+)
+
+// This code assembles the verdict structure, as expected by the
+// nftables netlink API.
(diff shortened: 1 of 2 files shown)


expr/verdict.go
```diff
@@ -0,0 +1,84 @@
...
+//  - netfilter.h (Linux kernel)
+//  - net/netfilter/nf_tables_api.c (Linux kernel)
+//  - src/expr/data_reg.c (linbnftnl)
+
+type Verdict uint64
+
+// Verdicts, as per netfilter.h.
+const (
+	Drop   Verdict = 0x00
+	Accept Verdict = 0x01
```
stapelberg: You’re missing Stolen (2). Could you add a Verdict prefix (so as to avoid clashes with other types) and change this to use iota while at it?

```go
const (
  VerdictDrop Verdict = iota
  VerdictAccept
  VerdictStolen
  // …
)
```

stapelberg: Also, we should get these constants into golang.org/x/sys/unix, but one step at a time :)

expr/verdict.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// This code assembles the verdict structure, as expected by the
+// nftables netlink API.
+// For further information, consult:
+//  - netfilter.h (Linux kernel)
+//  - net/netfilter/nf_tables_api.c (Linux kernel)
+//  - src/expr/data_reg.c (linbnftnl)
+
+type Verdict uint64
```
stapelberg: Are we sure that Verdict will never need to hold anything but a uint64? Might be prudent to use a struct so that we can extend it later.

expr/verdict.go
stapelberg: I’m not sure, but would it hurt to provide it? Maybe it will be used at some point. The C headers have it, so I think it should be available in Go.

stapelberg:
Also, please squash your changes so that we have nice logical commits in the git history in the end :)

stapelberg:
Thanks!

PR #8, opened 2019-01-27
Implement Queue Expr

Dhiver:
I needed to enqueue packet to a userspace application with netfilter queue so I added the Queue expression.

You can now use the queue statements listed [here](https://wiki.nftables.org/wiki-nftables/index.php/Quick_reference-nftables_in_10_minutes#Queue).

### DIFF
--- a/expr/queue.go
+++ b/expr/queue.go
@@ -1,0 +1,82 @@
+// Copyright 2018 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


Dhiver:
I signed it!

expr/queue.go
```diff
@@ -0,0 +1,68 @@
...
+	"golang.org/x/sys/unix"
+)
+
+type QueueAttribute uint16
+
+type QueueFlag uint16
+
+// Possible QueueAttribute values
+const (
+	QueueAttributeNum   QueueAttribute = unix.NFTA_QUEUE_NUM
```
stapelberg: For consistency, these should just be `QueueNum`, `QueueTotal` and `QueueFlags` (without the `Attribute` part).

expr/queue.go
```diff
@@ -0,0 +1,68 @@
+package expr
```
stapelberg: Could you copy the copyright header into this file please?

expr/queue.go
Dhiver: It's done.

expr/queue.go
Dhiver: It's done.

stapelberg:
Thanks for making the changes. Before I merge, could you squash your commits into one logical commit please, so that we can retain a clean git history?

Thanks!

Dhiver:
It's done :)

stapelberg:
Thanks!

PR #9, opened 2019-01-29
Implement support for nftable sets.

twitchy-jsonp:
Yo yo!!! *taps mic* How **you** doin? :P

Sooooooooo sets are really annoying and only seem to be documented in code. I've started implementing them (this is going to be a biiiiiig effort to get them full implemented) and I want your feedback on the API and the package layout.

My hope is to take that feedback and get anonymous sets working well, then do named sets in a separate PR. I don't intend to implement verdict sets.

And yes, I'll squash my commits once the PR is ready XD

### DIFF
--- a/nftables.go
+++ b/nftables.go
@@ -313,6 +313,7 @@ type Table struct {
 // Commands are buffered. Flush sends all buffered commands in a single batch.
 type Conn struct {
 	TestDial nltest.Func // for testing only; passed to nltest.Dial
+	NetNS    int         // Network namespace netlink will interact with.
 	messages []netlink.Message
 	err      error
 }
@@ -366,13 +367,7 @@ func (cc *Conn) Flush() error {
 	if cc.err != nil {
 		return cc.err // serialization error
 	}
-	var conn *netlink.Conn
-	var err error
-	if cc.TestDial == nil {
-		conn, err = netlink.Dial(unix.NETLINK_NETFILTER, nil)
-	} else {
-		conn = nltest.Dial(cc.TestDial)
-	}
+	conn, err := cc.dialNetlink()
 	if err != nil {
 		return err
 	}
@@ -394,17 +389,10 @@ func (cc *Conn) Flush() error {
 
 // GetRule returns the rules in the specified table and chain.
 func (cc *Conn) GetRule(t *Table, c *Chain) ([]*Rule, error) {
-	var conn *netlink.Conn
-	var err error
-	if cc.TestDial == nil {
-		conn, err = netlink.Dial(unix.NETLINK_NETFILTER, nil)
-	} else {
-		conn = nltest.Dial(cc.TestDial)
-	}
+	conn, err := cc.dialNetlink()
 	if err != nil {
 		return nil, err
 	}
-
 	defer conn.Close()
 
 	data, err := netlink.MarshalAttributes([]netlink.Attribute{
@@ -566,18 +554,18 @@ func objFromMsg(msg netlink.Message) (Obj, error) {
 	return nil, fmt.Errorf("malformed stateful object")
 }
 
-func (cc *Conn) getObj(o Obj, msgType uint16) ([]Obj, error) {
-	var conn *netlink.Conn
-	var err error
-	if cc.TestDial == nil {
-		conn, err = netlink.Dial(unix.NETLINK_NETFILTER, nil)
-	} else {
-		conn = nltest.Dial(cc.TestDial)
+func (cc *Conn) dialNetlink() (*netlink.Conn, error) {
+	if cc.TestDial != nil {
+		return nltest.Dial(cc.TestDial), nil
 	}
+	return netlink.Dial(unix.NETLINK_NETFILTER, &netlink.Config{NetNS: cc.NetNS})
+}
+
+func (cc *Conn) getObj(o Obj, msgType uint16) ([]Obj, error) {
+	conn, err := cc.dialNetlink()
 	if err != nil {
 		return nil, err
 	}
-
 	defer conn.Close()
 
 	data, err := o.marshal(false)
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -731,3 +731,101 @@ func TestConfigureClamping(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestDropVerdict(t *testing.T) {
+	// The want byte sequences come from stracing nft(8), e.g.:
+	// strace -f -v -x -s 2048 -eraw=sendto nft add table ip nat
+	//
+	// The nft(8) command sequence was taken from:
+	// https://wiki.nftables.org/wiki-nftables/index.php/Mangle_TCP_options
+	want := [][]byte{
+		// batch begin
+		[]byte("\x00\x00\x00\x0a"),
+		// nft flush ruleset
+		[]byte("\x00\x00\x00\x00"),
+		// nft add table ip filter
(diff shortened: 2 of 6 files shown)


set.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// SetElement represents a data point within a set.
+type SetElement struct {
+	Key []byte
+	Val []byte
+}
+
+// SetVal applies data points to an nftables set.
+func (cc *Conn) SetVal(s *Set, vals []SetElement) {
```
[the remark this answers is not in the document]
twitchyliquid64: Sure. Are you okay with me deviating from nft on stuff like this (and hence the tests changing)?

Also, LGTM on the `SetElement` structure?

twitchyliquid64:
I've just done a major refactor based on your comments and further testing, PTAL.

As testing exclusively with byte streams is no longer tenable, ive introduced a flag to enable tests that run against the live kernel. All the features I introduce that deviate from nftables will follow this approach for testing.

Ive now go two tests; the anonymous set test (which is byte identical with `nft`) and named sets (which is made against the netlink API directly).

nftables has a bunch of magic values that it uses to compute the fancy print output of `nft list ...`, which it puts in `USERDATA` and `SET_KEY_TYPE`. I've replicated this so that using the nft command will work as expected.

twitchyliquid64:
Heyaaaaaa, I think this is ready for review 4 real. Could you take a look?

Main thing ive done is added Delete/Get methods and tests for everything.

These tests flush/mutate the systems tables, but are now fully automatic.

nftables_test.go
```diff
@@ -28,6 +30,15 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+var (
+	enableSysTests = flag.Bool("run_system_tests", false, "Run tests that operate against the live kernel")
```
stapelberg: Could we run the test in a new network namespace to isolate its changes from the host system? See e.g. https://github.com/rtr7/router7/blob/cbff6a25413f53d87187f0a2cbb76a380f6bf28e/integration/netconfig/netconfig_test.go#L290

twitchy-jsonp: At the expense of some more dependencies (specifically github.com/vishvananda/netlink & github.com/vishvananda/netns), I can do this all in Go (and not have to call out to `ip` command). Should I do that instead?

(Also TIL rulesets are namespace aware, lol)

stapelberg: As long as these remain test-only dependencies, sure.

twitchy-jsonp: Doneskies!

twitchy-jsonp:
Done! Lemme know when you want me to squish my commits into merge juice, or if there are more changes.

twitchy-jsonp:
I've pre-emptively squashed commits.

nftables_test.go
```diff
@@ -74,6 +81,44 @@ func ifname(n string) []byte {
...
+// cleanupSystemNFTConn() must be called from a defer to cleanup
+// side effects of this function.
+func openSystemNFTConn(t *testing.T) (*nftables.Conn, netns.NsHandle, netns.NsHandle) {
+	t.Helper()
+	if !*enableSysTests {
+		t.SkipNow()
+	}
+	runtime.LockOSThread()
+
+	// Get the current namespace, so we can switch
```
stapelberg: Instead of the rather elaborate switching back and forth, could you please refactor the dial calls (e.g. https://github.com/google/nftables/blob/ce006eb26de8c3158d9d43bfee697c9aac5208a3/nftables.go#L572-L575) into a helper function which specifies a https://godoc.org/github.com/mdlayher/netlink#Config with NetNS set?

nftables_test.go
```diff
@@ -16,18 +16,25 @@ package nftables_test
...
 	"github.com/google/nftables"
 	"github.com/google/nftables/binaryutil"
 	"github.com/google/nftables/expr"
 	"github.com/mdlayher/netlink"
+	"github.com/vishvananda/netns"
 	"golang.org/x/sys/unix"
 )
 
+var (
+	enableSysTests = flag.Bool("run_system_tests", false, "Run tests that operate against the live kernel")
```
stapelberg: Now that we’re running these in a throwaway network namespace, is there any point in having the flag? Are there any adverse side-effects to protect developers from? If no, please remove.

twitchyliquid64: Yes, this is still too dangerous to let it run without active acknowledgment (though more IPTables fault than NFTables).

Here is one of many scenarios where we would fuck up the system:
1. User uses iptables on their system
2. User runs our tests
3. Our tests test the NAT functionality, causing the `nf_nat` kernel module to be loaded, along with a bunch others.
4. The load of `nf_nat` silently uses the (single) hook in the kernel for NAT functionality, silently breaking IPTables.

twitchyliquid64: I just tested without the `Flushruleset` bit and I can see some test remnants from _outside_ the network namespace - suggesting nftables is not fully namespace aware.

twitchyliquid64: Nevermind - forgot a single method of `Dial` in the refactor. Works fine in a network namespace.

nftables_test.go
twitchyliquid64: Sure thing. Given that running this in a network namespace could be a useful feature anyway, ive added the `NetNS` field to allow users to specify a network namespace.

twitchy-jsonp: I've left the thread locking as IIRC unshare syscall is a task-level operation. Could be wrong thou.

twitchy-jsonp:
PTAL! Fingers crossed this is the golden ticket :unicorn: :+1:

nftables.go
```diff
@@ -566,18 +554,23 @@ func objFromMsg(msg netlink.Message) (Obj, error) {
...
-		conn, err = netlink.Dial(unix.NETLINK_NETFILTER, nil)
-	} else {
-		conn = nltest.Dial(cc.TestDial)
+func (cc *Conn) dialNetlink() (*netlink.Conn, error) {
+	if cc.TestDial != nil {
+		return nltest.Dial(cc.TestDial), nil
+	}
+
+	var conf *netlink.Config
+	if cc.NetNS != 0 {
```
stapelberg: As 0 is the zero value of the NetNS field, you can skip this conditional:
```go
return netlink.Dial(unix.NETLINK_NETFILTER, &netlink.Config{NetNS: cc.NetNS})
```

nftables_test.go
stapelberg: Good point!

In that case, can you make sure that travis actually runs these tests, though?

nftables_test.go
```diff
@@ -74,6 +81,32 @@ func ifname(n string) []byte {
...
+// openSystemNFTConn returns a netlink connection that tests against
+// the running kernel in a separate network namespace.
+// cleanupSystemNFTConn() must be called from a defer to cleanup
+// created network namespace.
+func openSystemNFTConn(t *testing.T) (*nftables.Conn, netns.NsHandle) {
+	t.Helper()
+	if !*enableSysTests {
+		t.SkipNow()
+	}
+	runtime.LockOSThread()
```
stapelberg: Can you please add a comment stating that netns.New calls `unshare(2)` under the covers, which operates on the current thread?

twitchy-jsonp: Doneskies :]

nftables.go
twitchy-jsonp: Doneskies :]

twitchyliquid64:
Doneskies, woo!

I've left `go test ./...` in the travis config just in case other subpackages with tests appear later on.

stapelberg:
Great work on this, thanks! Merged.

PR #12, opened 2019-02-18
Readme

twitchy-jsonp:
This is just a draft - I havent made the table documenting all the different expressions and how to use them yet. Let me know if I am on the right track, I'll finish off this PR if yes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,78 @@ implemented in pure Go, i.e. does not wrap libnftnl.
 
 This is not an official Google product.
 
-## Breaking changes
+## Alpha status
 
-This package is in very early stages, and only contains enough data types and
-functions to install very basic nftables rules. It is likely that mistakes with
-the data types/API will be identified as more functionality is added.
+This package is in early stages, and only implements a subset of nftables
+features. While the developers intend to keep interfaces & function signatures
+backwards-compatible, no guarantees are made; bugs or any unexpected
+structuring of nftables features may result in breaking changes.
+
+## Usage
+
+Issue commands to mutate or read nftables state. Commands that mutate state
+(eg: `AddTable`, `AddChain`, `AddSet`, `AddRule`) are queued until `Flush`
+is called.
+
+### Expressions
+
+The following expressions are implemented for use in rule logic:
+
+TODO
+
+### Examples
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -6,11 +6,78 @@ implemented in pure Go, i.e. does not wrap libnftnl.
...
+
+Issue commands to mutate or read nftables state. Commands that mutate state
+(eg: `AddTable`, `AddChain`, `AddSet`, `AddRule`) are queued until `Flush`
+is called.
+
+### Expressions
+
+The following expressions are implemented for use in rule logic:
+
+TODO
```
stapelberg: I think we should just link to godoc instead of maintaining a separate list. It will just go stale.

README.md
```diff
@@ -6,11 +6,78 @@ implemented in pure Go, i.e. does not wrap libnftnl.
...
+
+### Examples
+
+#### Drop outgoing packets to 1.2.3.4
+
+```go
```
stapelberg: Could we link to a compiled example in godoc instead? That way, it’ll always be up to date.

README.md
```diff
@@ -6,11 +6,78 @@ implemented in pure Go, i.e. does not wrap libnftnl.
 
 This is not an official Google product.
 
-## Breaking changes
+## Alpha status
```
stapelberg: Can you keep the “Breaking changes” in the heading please? I’d like to have it prominently pointed out.

README.md
twitchyliquid64: I'm not a huge fan of long examples in godoc, could we have an 'examples' folder with a bunch instead?

Alternatively, I'm writing up heaps with this package on [my](https://tomdnetto.net/post/linux_networking_from_go_nftables) [blog](https://tomdnetto.net/post/advanced_nftables_with_go). I don't think it would be right to link to my blog, but maybe we could put equivalent content in a wiki on this repo?

stapelberg: An examples directory works, too, but please make them be testable.

I don’t mind linking to your blog (in addition). That might be a more pragmatic alternative than wikis, which only make sense for larger projects where there’s enough motivation around to keep them up-to-date.

PR #13, opened 2019-03-02
nftables: update for package netlink constant renames

mdlayher:
We renamed these constants in https://github.com/mdlayher/netlink/pull/127 to remove the prefix.


stapelberg:
Thanks!

PR #15, opened 2019-05-01
Restructure code base into smaller files

rwhelan:
I found reasoning about the code a little cumbersome while evaluating functionality with everything in a single file.   Hopefully this makes it easier to find relevant sections while looking through the code?  I changed almost nothing (besides file layout) with the exception of: 

- Removal of the function `stringFrom0`;  I could not find it used anywhere in the code
- Relocating the function `exprsFromMsg` into the sub package `expr`

:man_shrugging:


stapelberg:
Sorry for the delay, I’m currently traveling.

The code reorganization is fine per se, but:

> * Relocating the function `exprsFromMsg` into the sub package `expr`

This expands the public API surface. Are you planning to use ExprsFromMsg for something (if so, what)? Or is that just an artifact of moving the code around. If the latter, can you revert that part of the change please?

Also, I noticed that with your PR applied, godoc no longer shows the package-level comment on package nftables. Can you restore that please?

rwhelan:
At 11:30 at night when I did this, it felt like expr was the right place for the `exprsFromMsg` function to be.  I've moved it to the `rule.go` file where its used.

I created a file called doc.go for the package level comment.  If another little file is to much clutter, it can be moved.

Thanks for looking at this!

PR #16, opened 2019-05-01
List tables

rwhelan:
This PR adds `ListTables()` to retrieve currently configured tables within the kernel.

This PR builds on PR#15

### DIFF
--- a/table.go
+++ b/table.go
@@ -15,17 +15,22 @@
 package nftables
 
 import (
+	"fmt"
+
 	"github.com/mdlayher/netlink"
 	"golang.org/x/sys/unix"
 )
+
+var tableHeaderType = netlink.HeaderType((unix.NFNL_SUBSYS_NFTABLES << 8) | unix.NFT_MSG_NEWTABLE)
 
 // TableFamily specifies the address family for this table.
 type TableFamily byte
 
 // Possible TableFamily values.
 const (
-	TableFamilyIPv4 TableFamily = unix.AF_INET
-	TableFamilyIPv6 TableFamily = unix.AF_INET6
+	TableFamilyIPv4   TableFamily = unix.AF_INET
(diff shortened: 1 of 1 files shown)


stapelberg:
Could you rebase this please?

RyanAtCNG:
Of Course!

Other List functions my be on the horizon as my project matures.

Thanks!

rwhelan:
Of Course!

More List functions maybe on the way as my project matures.

Thanks for the library; very helpful!

table.go
```diff
@@ -66,3 +71,66 @@ func (cc *Conn) AddTable(t *Table) *Table {
...
+	ad, err := netlink.NewAttributeDecoder(msg.Data[4:])
+	if err != nil {
+		return nil, err
+	}
+
+	for ad.Next() {
+		switch ad.Type() {
+		case unix.NFTA_TABLE_NAME:
+			t.Name = ad.String()
+			/*
```
stapelberg: What about this commented out code? Is there anything left to do here (if so, can you implement it?) or is this cruft?

rwhelan: They are attributes returned from the kernel that are not yet represented on the Table struct.  I found them when decoding kernel responses- but don't yet have a need for them.  I put them (commented out) on the decoding function as a reminder but forgot they were there when submitting the PR.

I will remove them for now; unless you'd prefer to extend the structs to represent the attributes for completeness?

stapelberg: If they’re easy enough to test, let’s do completeness to save the person needing them some work :) 

If not, you can leave them out.

rwhelan:
I added 2 of the struct members to the decoding and everything looks to be working.  I contemplated created an enum for the `Flags` values, but looking at the header file on my machine, it looks like there is only a single flag defined...

stapelberg:
Thanks!

Issue #61, opened 2019-10-15
Implement value setting for ct


Minaru:
Currently we are only able to read information from the conntrack module of nftables through this package.

The goal would be to be able to write into it, for example to set the conntrack mark or other keys.

The base code is already working in the Meta module, it just has to be taken and adapted for ct with the proper DREG/SREG related to NFTA_CT.

I'll make a PR as soon as the code is ready, it shouldn't take too long as it's only about adding a field in the struct/modifying (un)marshal functions.

Minaru:
Closing this as this is merged.

PR #63, opened 2019-10-15
Set ct values

Minaru:
Reference issue #61

Took exactly the `Meta` marshal function (which already has source/dest reg mechanism implemented) and implemented it here. ([ref](https://github.com/google/nftables/blob/master/expr/expr.go#L83))

`CtKey` values are taken from [unix golang package](https://godoc.org/golang.org/x/sys/unix#NFT_CT_STATE).

Awaiting your feedback !

### DIFF
--- a/expr/ct.go
+++ b/expr/ct.go
@@ -22,6 +22,33 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// CtKey specifies which piece of conntrack information should be loaded. See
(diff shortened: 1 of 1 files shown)


Minaru:
@googlebot I signed it!

Minaru:
I'll remake one properly as this one is getting messy (been a long time I haven't done a PR)

PR #92, opened 2020-01-14
Fix obj api

alexispires:
This PR fix the Obj API:

Currently:
- GetObj(Obj) : return all counters, not only the given one 
- GetObjReset(Obj) : reset all counters, not only the given one

The consequence of that:
- No clean API to retrieve all counters
- No API to retrieve values of a given counter
- No API to reset a given counter

This PR add the following methods:
GetObj(o

-  Obj) (Obj, error)
- GetObjs(t Table) ([]Obj, error)
- ResetObj(o Obj) (Obj, error)
- ResetObjs(t Table) ([]Obj, error)

Related to https://github.com/google/nftables/issues/90

### DIFF
--- a/obj.go
+++ b/obj.go
@@ -66,7 +66,6 @@ func (cc *Conn) GetObjReset(o Obj) ([]Obj, error) {
 	return cc.getObj(nil, o.table(), unix.NFT_MSG_GETOBJ_RESET)
 }
 
-
 // GetObject gets the specified Object
 func (cc *Conn) GetObject(o Obj) (Obj, error) {
 	objs, err := cc.getObj(o, o.table(), unix.NFT_MSG_GETOBJ)
--- a/obj.go
+++ b/obj.go
@@ -54,23 +54,38 @@ func (cc *Conn) AddObj(o Obj) Obj {
 	return o
 }
 
-// GetObj gets the specified Obj without resetting it.
-func (cc *Conn) GetObj(o Obj) (Obj, error) {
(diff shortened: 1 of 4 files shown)


stapelberg:
Can you use a more detailed commit message please? The gist of what you reported in the issue should be represented in there

alexispires:
@stapelberg it's ok now?

stapelberg:
Commit title “fix obj api” is not descriptive enough: what exactly are you fixing? In the message, describe the behavior before your fix and after your fix.

alexispires:
It's better now? @stapelberg

obj.go
```diff
@@ -112,24 +123,35 @@ func objFromMsg(msg netlink.Message) (Obj, error) {
...
 	if err != nil {
 		return nil, err
 	}
 	defer conn.Close()
 
-	data, err := o.marshal(false)
-	if err != nil {
-		return nil, err
+	var data []byte
+	var message netlink.Message
```
stapelberg: Why are you moving the declaration of message up? This seems unnecessary

obj.go
```diff
@@ -54,13 +55,23 @@ func (cc *Conn) AddObj(o Obj) Obj {
 }
 
 // GetObj gets the specified Obj without resetting it.
-func (cc *Conn) GetObj(o Obj) ([]Obj, error) {
```
stapelberg: Can you leave the old API untouched and instead introduce a new function to avoid breaking the API? Perhaps `GetObject`, or `GetSingleObject`

PR #93, opened 2020-01-16
function to create concatenated SetDatatypes

LogicalOverflow:
I added an exported function to create easily concatenated SetDatatypes (and some tests for it).

### DIFF
--- a/set.go
+++ b/set.go
@@ -81,11 +81,25 @@ var (
 	}
 )
 
-// ConcatSetType constructs a new SetDatatype which consists of a concatenation of the passed types. It panics, if the
-// nftMagic would overflow (more than 5 types)
-func ConcatSetType(types ...SetDatatype) SetDatatype {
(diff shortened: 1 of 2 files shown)


set.go
```diff
@@ -77,6 +81,30 @@ var (
 	}
 )
 
+// ConcatSetType constructs a new SetDatatype which consists of a concatenation of the passed types. It panics, if the
+// nftMagic would overflow (more than 5 types)
+func ConcatSetType(types ...SetDatatype) SetDatatype {
+	if len(types) > 32/SetConcatTypeBits {
+		panic("too many type to concat")
```
stapelberg: Let’s return an error instead of panicing here

LogicalOverflow: Sure. I'll added a panicing variant (MustConcatSetType), too simplify safe initializations of global variables.

stapelberg:
Thanks!

PR #98, opened 2020-02-25
Report whether set has timeout flag configured

zacatac:
Per-element timeouts should be configurable for a set even if a top-level set timeout isn't configured.

Example:
```
> nft list map ip nat porttoip
table ip nat {
        map porttoip {
                type inet_service : ipv4_addr
                flags timeout
                elements = { 8000 timeout 1m expires 49s32ms : 100.0.0.0 }
        }
}
```


stapelberg:
Thanks!

PR #101, opened 2020-03-12
Add support for rate limiting

gregdel:
This PR adds support for rate limiting.

I'm not sure about the error handling in this PR, I created typed errors but it's not what's done in the other files of this lib. What do you think ?

### DIFF
--- a/expr/limit.go
+++ b/expr/limit.go
@@ -1,0 +1,136 @@
+// Copyright 2018 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package expr
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+
+	"github.com/google/nftables/binaryutil"
+	"github.com/mdlayher/netlink"
+	"golang.org/x/sys/unix"
+)
+
+// LimitType represents the type of the limit expression.
+type LimitType uint32
+
+// Imported from the nft_limit_type enum in netfilter/nf_tables.h.
+const (
+	LimitTypePkts     LimitType = unix.NFT_LIMIT_PKTS
+	LimitTypePktBytes LimitType = unix.NFT_LIMIT_PKT_BYTES
+)
+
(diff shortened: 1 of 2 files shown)


expr/limit.go
```diff
@@ -0,0 +1,153 @@
...
+	"errors"
+
+	"github.com/google/nftables/binaryutil"
+	"github.com/mdlayher/netlink"
+	"golang.org/x/sys/unix"
+)
+
+// Theses errors represent the possible parsing error of the rate limit.
+var (
+	ErrLimitTimeUnitInvalid = errors.New("expr: invalid limit time unit")
```
stapelberg: This is not actually used in the code, it seems? Remove?

expr/limit.go
```diff
@@ -0,0 +1,153 @@
...
+		return 0, errors.New("nftables: invalid limit value")
+	}
+}
+
+// Limit represents a rate limit expression
+type Limit struct {
+	Type  LimitType
+	Rate  uint64
+	Over  bool
+	Unit  LimitTime
```
stapelberg: Why not make Unit a time.Duration and convert it in marshal()? That would feel more familiar to users, I think.

gregdel:
Hey @stapelberg, thanks for your review.

I've taken your remarks into account:
* the errors are not typed anymore (I like typing errors to make the job of the caller easier if he needs to check the error type)
* the `Unit` parameter is now a `time.Duration`, I've added a comment about the possible values of this field

Don't hesitate if you have any other remarks.

expr/limit.go
```diff
@@ -0,0 +1,107 @@
...
+	LimitTypePktBytes LimitType = unix.NFT_LIMIT_PKT_BYTES
+)
+
+// Limit represents a rate limit expression.
+type Limit struct {
+	Type LimitType
+	Rate uint64
+	Over bool
+	// Possible values of unit:
+	// second, minute, hour, day, week
```
stapelberg: This is a little vague. Are you trying to say that the granularity of `Unit` is whole seconds, and the maximum representable value is 1 week?

gregdel: nft will display the rules only for a second, a minute, an hour, a day and a week. This is an enum in the nft source code 
e.g.
```
udp dport 5006 limit rate 10000 bytes/minute
ip saddr 109.190.254.33 tcp dport 8080 limit rate 10/second
```

Now that the type is a time.Duration, you could set the duration to `2 * time.Second`, in that case, the rule will apply but nft will display `bytes/error`. I'm not sure if it works as expected though.

What you be better for the user, a better comment or an "enum" of type time.Duration representing the possible values ?

stapelberg: Ah, now I understand. In that case, using time.Duration doesn’t make sense (users would need to convert/restrict them) after all, and I think we should just go for enum values with a custom type.

gregdel: I reintroduced the custom type with the possible values.

stapelberg:
Thanks!

Issue #105, opened 2020-08-02
bug: chain name is empty when unpacking expr.Verdict


greenpau:
The chain `FORWARD` looks as follows:

```
table ip6 filter {
        chain FORWARD {
                type filter hook forward priority filter; policy drop;
                jump cnins-3-4026600826-dummy0
        }

        chain cnins-3-4026600826-dummy0 {
                oifname "dummy0" ip6 daddr 2001:db8:100:100::1 ct state established,related counter packets 0 bytes 0 accept
                iifname "dummy0" ip6 saddr 2001:db8:100:100::1 counter packets 0 bytes 0 accept
                iifname "dummy0" oifname "dummy0" counter packets 0 bytes 0 accept
        }
}
```

The `jump` rules was created with:

```go
            // Finally, add the jump to the above chain in FORWARD chain.
            jumpRule := &nftables.Rule{
                Table: addr.table,
                Chain: addr.chain,
                Exprs: []expr.Any{},
            }
            jumpRule.Exprs = append(jumpRule.Exprs, &expr.Verdict{
                Kind:  expr.VerdictJump,
                Chain: chainName,
            })

            nb.conn.AddRule(jumpRule)
            if err := nb.conn.Flush(); err != nil {
                return fmt.Errorf(
                    "failed adding jump rule in table %s chain %s for address %v of interface %s",
                    addr.table.Name, addr.chain.Name, addr.conf, intfName,
                )
            }
```

Subsequently, when retriving the rule, it does not resolve to `expr.VerdictJump` and `Chain` name is empty:

```
([]*nftables.Rule) (len=1 cap=1) {
 (*nftables.Rule)(0xc00033ae60)({
  Table: (*nftables.Table)(0xc00000f400)({
   Name: (string) (len=6) "filter",
   Use: (uint32) 0,
   Flags: (uint32) 0,
   Family: (nftables.TableFamily) 0
  }),
  Chain: (*nftables.Chain)(0xc00045e780)({
   Name: (string) (len=7) "FORWARD",
   Table: (*nftables.Table)(<nil>),
   Hooknum: (nftables.ChainHook) 0,
   Priority: (nftables.ChainPriority) 0,
   Type: (nftables.ChainType) "",
   Policy: (*nftables.ChainPolicy)(<nil>)
  }),
  Position: (uint64) 0,
  Handle: (uint64) 6,
  Exprs: ([]expr.Any) (len=1 cap=1) {
   (*expr.Verdict)(0xc00000f580)({
[...]

PR #106, opened 2020-08-02
fix: unmarshaling verdicts with chain information

greenpau:
Before this commit: the unmarshaling of a verdict pointing
to a chain fails.

After this commit: the unmarshaling of a rule with a verdict
pointing to a chain succeeds and the information about the
chain gets put in `Verdict.Chain`.

Resolves: #105

Signed-off-by: Paul Greenberg <greenpau@outlook.com>

### DIFF
--- a/expr/verdict.go
+++ b/expr/verdict.go
@@ -15,6 +15,7 @@
 package expr
 
 import (
+	"bytes"
 	"encoding/binary"
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks!

greenpau:
@stapelberg , thank you! :-) love the fact you did it on weekend!!! Is it possible to tag and release? I guess it is go modules thing nowadays.

stapelberg:
We don’t have any tags right now, as compatibility-wise we don’t want to make too many promises yet.

That said, you don’t need to use tags to use Go modules, the only benefit in this case is that tags look a bit nicer than the auto-generated version identifiers.

Issue #104, opened 2020-08-01
feature: Rule methods


greenpau:
@stapelberg, @sbezverk and other maintainers, currently many module's structs have no methods of their own. For example, it would have been nice to have `String()` method for `Rule`. 

Is it OK to submit a PR for this?

stapelberg:
Regarding the `String()` method, just to double-check: it would return a human-readable representation of the rule, and wouldn’t be used programmatically, yes? If so, that seems okay to me.

greenpau:
> Regarding the String() method, just to double-check: it would return a human-readable representation of the rule, and wouldn’t be used programmatically, yes?

@stapelberg , Yes. I want to be able to see one liner rule, similar to “nft” utility. Today, it takes an “essay” to get a gist of what a rule does! 😄 i am not complaining ... i appreciate the complexity ... i learned from the experience ... however, it would probably make the library easier to consume. In fact, back in March, i failed to use the library because of its use complexity.  Then, I stumbled upon a few blog posts and had an epiphany :-)

greenpau:
It is just Rule does not have any method ... For example ... ‘IsVerdictDrop’, ‘IsVerdictJump’, ‘IsIPv4’, ‘IsIPv6’,  etc.

stapelberg:
Sure, in that case adding String() makes sense.

For the other methods, I’m tentatively in favor, but let’s discuss those over a pull request maybe?

In general, this library is pretty low-level, so I don’t want to add too much syntactic sugar.

greenpau:
> For the other methods, I’m tentatively in favor, but let’s discuss those over a pull request maybe?

@stapelberg , absolutely!

> In general, this library is pretty low-level, so I don’t want to add too much syntactic sugar.

👍

PR #108, opened 2020-08-03
WIP: rule: add String() method

greenpau:
Before this commit: the printing of a rule results in
a pointer address.

After this commit: the printing of a rules results in
a human-readable text.

Resolves: #104

Signed-off-by: Paul Greenberg <greenpau@outlook.com>

### DIFF
--- a/expr/verdict.go
+++ b/expr/verdict.go
@@ -24,6 +24,15 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+const (
+	NFT_DROP   = 0
+	NFT_ACCEPT = 1
+	NFT_STOLEN = 2
+	NFT_QUEUE  = 3
+	NFT_REPEAT = 4
+	NFT_STOP   = 5
+)
+
 // This code assembles the verdict structure, as expected by the
 // nftables netlink API.
 // For further information, consult:
@@ -125,4 +134,38 @@ func (e *Verdict) unmarshal(data []byte) error {
 		}
 	}
 	return ad.Err()
+}
+
+func (e *Verdict) String() string {
+	var v string
+	switch e.Kind {
+	case unix.NFT_RETURN:
+		v = "return" // -0x5
+	case unix.NFT_GOTO:
+		v = "goto" // -0x4
+	case unix.NFT_JUMP:
+		v = "jump" // NFT_JUMP = -0x3
+	case unix.NFT_BREAK:
+		v = "break" // NFT_BREAK = -0x2
+	case unix.NFT_CONTINUE:
+		v = "continue" // NFT_CONTINUE = -0x1
+	case NFT_DROP:
+		v = "drop"
+	case NFT_ACCEPT:
+		v = "accept"
+	case NFT_STOLEN:
+		v = "stolen"
+	case NFT_QUEUE:
+		v = "queue"
+	case NFT_REPEAT:
+		v = "repeat"
+	case NFT_STOP:
+		v = "stop"
+	default:
+		v = fmt.Sprintf("verdict %v", e.Kind)
+	}
+	if e.Chain != "" {
+		return v + " " + e.Chain
+	}
+	return v
 }
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -252,12 +252,27 @@ func TestRuleOperations(t *testing.T) {
 		expr.VerdictDrop,
 	}
 
+	wantStrings := []string{
+		"queue",
+		"accept",
+		"queue",
+		"accept",
+		"drop",
+		"drop",
+	}
+
 	for i, r := range rules {
 		rr, _ := r.Exprs[0].(*expr.Verdict)
 
 		if rr.Kind != want[i] {
 			t.Fatalf("bad verdict kind at %d", i)
 		}
+
+		if rr.String() != wantStrings[i] {
+			t.Fatalf("bad verdict string at %d: %s (received) vs. %s (expected)", i, rr.String(), wantStrings[i])
+		}
+
+		t.Logf("%s", rr)
 	}
 }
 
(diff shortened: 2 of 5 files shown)


expr/verdict.go
```diff
@@ -24,6 +24,15 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+const (
+	NFT_DROP   = 0
+	NFT_ACCEPT = 1
+	NFT_STOLEN = 2
+	NFT_QUEUE  = 3
+	NFT_REPEAT = 4
+	NFT_STOP   = 5
```
stapelberg, line 33: Why did we not have these declared before? Where are they canonically defined in the upstream sources? Can you add a link please?

expr/verdict.go
```diff
@@ -126,3 +135,37 @@ func (e *Verdict) unmarshal(data []byte) error {
 	}
 	return ad.Err()
 }
+
+func (e *Verdict) String() string {
+	var v string
+	switch e.Kind {
+	case unix.NFT_RETURN:
+		v = "return" // -0x5
```
stapelberg, line 143: What did you want to express with these comments?

nftables_test.go
```diff
@@ -252,12 +252,27 @@ func TestRuleOperations(t *testing.T) {
...
 
 		if rr.Kind != want[i] {
 			t.Fatalf("bad verdict kind at %d", i)
 		}
+
+		if rr.String() != wantStrings[i] {
+			t.Fatalf("bad verdict string at %d: %s (received) vs. %s (expected)", i, rr.String(), wantStrings[i])
+		}
+
+		t.Logf("%s", rr)
```
stapelberg, line 275: remove debugging left-over?

nftables_test.sh
```diff
@@ -0,0 +1,3 @@
+go test ./...
```
stapelberg, line 1: Can you remove this file from the commit?

Making testing a little easier is a good idea in general, but I’d like to avoid establishing precedent of shell scripts :)

Maybe a “make test” target in a Makefile in a separate PR would be the best way forward?

expr/verdict.go
greenpau: > Why did we not have these declared before? Where are they canonically defined in the upstream sources? Can you add a link please?

There is no definition. They are defined with the following:

```
const (
    VerdictReturn VerdictKind = iota - 5
    VerdictGoto
    VerdictJump
    VerdictBreak
    VerdictContinue
    VerdictDrop
    VerdictAccept
    VerdictStolen
    VerdictQueue
    VerdictRepeat
    VerdictStop
)
```

I am looking in https://github.com/torvalds/linux/blob/47ec5303d73ea344e84f46660fff693c57641386/include/uapi/linux/netfilter/nf_tables.h#L64-L70 ... yet to discover.

expr/verdict.go
greenpau: > What did you want to express with these comments?

@stapelberg , just a note as to what the id is for the return.

nftables_test.go
greenpau: > remove debugging left-over?

@stapelberg , 👍 will do prior to removing WIP.

nftables_test.sh
greenpau: > Can you remove this file from the commit?

@stapelberg , 👍  will do prior to removing WIP.

> Maybe a “make test” target in a Makefile in a separate PR would be the best way forward?

@stapelberg , that would be nice! I agree.

expr/verdict.go
greenpau: another reference: https://github.com/torvalds/linux/blob/b71acb0e372160167bf6d5500b88b30b52ccef6e/include/uapi/linux/netfilter.h#L10-L17

greenpau: @stapelberg , in retrospect, after finding https://git.netfilter.org/libnftnl/tree/src/utils.c#n182, I say was mimicking this function 😄 

```c
const char *nftnl_verdict2str(uint32_t verdict)
{
	switch (verdict) {
	case NF_ACCEPT:
		return "accept";
	case NF_DROP:
		return "drop";
	case NF_STOLEN:
		return "stolen";
	case NF_QUEUE:
		return "queue";
	case NF_REPEAT:
		return "repeat";
	case NF_STOP:
		return "stop";
	case NFT_RETURN:
		return "return";
	case NFT_JUMP:
		return "jump";
	case NFT_GOTO:
		return "goto";
	case NFT_CONTINUE:
		return "continue";
	case NFT_BREAK:
		return "break";
	default:
		return "unknown";
	}
}
```

greenpau: @stapelberg , also the NF_STOLEN comes from https://git.netfilter.org/libnftnl/tree/include/linux/netfilter.h#n9

```c
/* Responses from hook functions. */
#define NF_DROP 0
#define NF_ACCEPT 1
#define NF_STOLEN 2
#define NF_QUEUE 3
#define NF_REPEAT 4
#define NF_STOP 5
#define NF_MAX_VERDICT NF_STOP
```

stapelberg: After having another look, I think this duplicates the verdict codes we already have defined here, no? https://github.com/google/nftables/blob/c25e4f69b4251e2e4175e5407a4ced5682099453/expr/verdict.go#L39-L54

Why not just use e.g. VerdictReturn in the switch statement below in String()?

expr/verdict.go
stapelberg: If you follow my other suggestion above regarding using the `Verdict*` names, I think it’d be cleaner to not duplicate these values (we only want a single place to update, not multiple). If people are inclined to find out the value, they should jump to the definition using their editor.

stapelberg (changes requested):
Sorry for the delay. I’ll try to respond more quickly on this PR.

glaslos:
@greenpau are you still planning to get this PR merged? Just started playing with this library and your branch was super helpful for debugging 🙏

PR #124, opened 2021-06-01
Added Other expression type to support types that are not explicitly supported

mlitvin:
Currently those attribute are silently ignored, so there is not even a clear indication that something went wrong.


thediveo:
@mlitvin this seems to be stuck? Maybe it needs the at-googlebot message?

Issue #114, opened 2020-08-29
bug: GetRule returns entries for all chains, not the one provided


greenpau:
I make `GetRule()` call in the following way.

```go
	tb := &nftables.Table{
		Name: tableName,
	}
	if v == "4" {
		tb.Family = nftables.TableFamilyIPv4
	} else {
		tb.Family = nftables.TableFamilyIPv6
	}

	ch := &nftables.Chain{
		Name:  chainName,
		Table: tb,
	}

	rules, err := conn.GetRule(tb, ch)
	if err != nil {
		return nil, err
	}
```

The issue is I am getting back all chains, not the table and change I passed as arguments.
Here is a dump of such response, Please note the `Name: (string) (len=10) "prerouting"` and `Name: (string) (len=11) "postrouting",`.

```
(*utils.chainInfo)(0xc00005d6d0)({
         RuleCount: (int) 9,
         Positions: ([]uint64) (len=9 cap=16) {
          (uint64) 0,
          (uint64) 0,
          (uint64) 0,
          (uint64) 7,
          (uint64) 8,
          (uint64) 0,
          (uint64) 0,
          (uint64) 0,
          (uint64) 2
         },
         Handles: ([]uint64) (len=9 cap=16) {
          (uint64) 6,
          (uint64) 11,
          (uint64) 7,
          (uint64) 8,
          (uint64) 9,
          (uint64) 12,
          (uint64) 2,
          (uint64) 2,
          (uint64) 3
         },
         Rules: ([]*nftables.Rule) (len=9 cap=16) {
          (*nftables.Rule)(0xc00005d770)({
           Table: (*nftables.Table)(0xc0001dc8e0)({
            Name: (string) (len=3) "nat",
            Use: (uint32) 0,
            Flags: (uint32) 0,
            Family: (nftables.TableFamily) 0
           }),
           Chain: (*nftables.Chain)(0xc0001cad80)({
            Name: (string) (len=11) "postrouting",
            Table: (*nftables.Table)(<nil>),
            Hooknum: (nftables.ChainHook) 0,
            Priority: (nftables.ChainPriority) 0,
            Type: (nftables.ChainType) "",
            Policy: (*nftables.ChainPolicy)(<nil>)
           }),
           Position: (uint64) 0,
           Handle: (uint64) 6,
           Exprs: ([]expr.Any) (len=1 cap=1) {
            (*expr.Verdict)(0xc0001dc9e0)({
             Kind: (expr.VerdictKind) -3,
[...]

PR #133, opened 2021-11-17
[FIX]fix GetRule method https://github.com/google/nftables/issues/114

RandolphCYG:
and add GetTable and GetChain method to get specific data by name && add test files


rule.go
```diff
@@ -84,7 +84,10 @@ func (cc *Conn) GetRule(t *Table, c *Chain) ([]*Rule, error) {
 		if err != nil {
 			return nil, err
 		}
-		rules = append(rules, r)
+
+		if r.Table.Name == t.Name && r.Table.Family == t.Family && r.Chain.Name == c.Name {
```
stapelberg, line 88: If I read https://elixir.bootlin.com/linux/v5.15.2/source/net/netfilter/nf_tables_api.c#L3136 correctly, I think setting the NLM_F_ECHO flag (which we do in line 68) results in dumping all rules regardless of the arguments.

So, the better fix might be to not set NLM_F_ECHO, so that the kernel only returns the data we are requesting?

tests/get_data_test.go
```diff
@@ -0,0 +1,52 @@
+package tests
```
stapelberg, line 1: We don’t have a separate tests package in the nftables package. 

These tests look different than the ones we already have. I think yours might not work on the GitHub Actions Continuous Integration as they are.

stapelberg (changes requested):
It seems like you’re changing file permissions of all files from 644 to 755 for some reason? Maybe an accidental misconfiguration on your end.

tests/get_data_test.go
RandolphCYG: > It seems like you’re changing file permissions of all files from 644 to 755 for some reason? Maybe an accidental misconfiguration on your end.

Because the file permission is set to 644, my Goland cannot open the project, so I changed it to 777. I re-decided on a better method, thanks for reminding

rule.go
RandolphCYG: Yes, your analysis is correct. At first, I only considered the frame selection of data at this layer, but ignored the restrictions when the c language api was first adjusted. Testing is also possible.

tests/get_data_test.go
RandolphCYG: Regarding the test file, I will confidently study the nftables_test file in the project and add it here. In addition, I hope the README file can provide more cases for users to know how to get started quickly

PR #135, opened 2021-11-22
[FIX]fix GetRule method https://github.com/google/nftables/issues/114…

randolphcyg:
… && add GetTable GetChain method


stapelberg:
The tests don’t seem to pass?
```
--- FAIL: TestGetTable (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x51aeae]
```

BTW, you don’t need to create a new pull request each time you change something, you can update an existing pull request: https://stackoverflow.com/questions/9790448/how-to-update-a-pull-request-from-forked-repo

RandolphCYG:
> 

sorry, I just added a part of the test case, I will fill it up tomorrow, and today I am still testing complex named set operations

RandolphCYG:
hey, could you help me check it again? tks

chain.go
```diff
@@ -205,6 +205,41 @@ func (cc *Conn) ListChains() ([]*Chain, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, m := range response {
+		c, err := chainFromMsg(m)
+		if err != nil {
+			return nil, err
+		}
+		if c.Name == name {
```
stapelberg, line 234: Here and for GetTable(), you’re still requesting all chains/tables from the kernel, and then filter after receiving them.

As discussed previously, it seems better to me to only request the chains/tables you need.

You’re already not setting NLM_F_ECHO, so perhaps all that’s needed is to specify the name/family attributes when sending the request to the kernel? Have you tried that?

nftables_test.go
```diff
@@ -113,6 +113,49 @@ func cleanupSystemNFTConn(t *testing.T, newNS netns.NsHandle) {
...
 }
 
+func TestGetTableGetChain(t *testing.T) {
+	conn := nftables.Conn{}
+	nat2 := conn.AddTable(&nftables.Table{
+		Family: nftables.TableFamilyIPv4,
+		Name:   "nat2",
+	})
+
+	fmt.Println(nat2.Name, nat2.Family)
```
stapelberg, line 123: Here and below, remove fmt.Print* messages that are left-over from debugging.

stapelberg (changes requested):

chain.go
RandolphCYG: ok, i will have a try, nice suggestions

PR #137, opened 2021-12-09
nftables: fix staticcheck error for Conn.getObj

mdlayher:
Signed-off-by: Matt Layher <mdlayher@gmail.com>

The error was not checked in the else case, so move the error check below the conditional.


stapelberg (approved):

stapelberg:
Thanks!

Issue #126, opened 2021-08-23
Adding a rule on position (index) 0


constcrist:
I have not checked yet the netlink/netfilter code in the linux kernel for adding a new nf rule.
however, based on the `nft(8)`, "RULES" section, a `0` value for index is valid:
```
The add and insert commands support an optional location specifier, 
which is either a handle or the index (starting at zero) of an existing rule.
```
as far as I understand from the code, `Rule.Position` is basically the `index` documented in the `nft(8)` man page.
currently, the code in `rule.go` treats the `0` `Rule.Position` as "don't care" index:
```
    if r.Position != 0 {
        msgData = append(msgData, cc.marshalAttr([]netlink.Attribute{
            {Type: unix.NFTA_RULE_POSITION, Data: binaryutil.BigEndian.PutUint64(r.Position)},
        })...)
    }
```
is there any reason for this?
imo, if `0` is indeed a valid index, there should be a flag added to the `Rule` data structures which signals whether `Rule.Position` should be used or not.

stapelberg:
cc @sbezverk 

> is there any reason for this?

Seems like an oversight to me

> imo, if `0` is indeed a valid index, there should be a flag added to the `Rule` data structures which signals whether `Rule.Position` should be used or not.

Agreed

PR #143, opened 2022-01-19
Support for rule position 0

turekt:
Hi all,

I have added a change that resolves the issue #126.

I checked the original netfilter and libnftnl source code and found that the `nftnl_rule` struct [contains a `uint32_t flags`](https://git.netfilter.org/libnftnl/tree/include/rule.h#n7) which is checked [during payload build](https://git.netfilter.org/libnftnl/tree/src/rule.c#n294). Some of the [`nftnl_rule_attr` flags](https://git.netfilter.org/libnftnl/tree/include/libnftnl/rule.h#n21) can be found in the [go unix package](https://cs.opensource.google/go/x/sys/+/da31bd32:unix/ztypes_linux.go;l=1894) with correct values. Therefore, my approach was to add a `Flags uint32` to the `Rule` struct and check for the `unix.NFTA_RULE_POSITION` bit. The `r.Position != 0` check remains because otherwise the change would not be backwards compatible.

Let me know what you think.

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -42,6 +42,11 @@ type Rule struct {
 	Chain    *Chain
 	Position uint64
 	Handle   uint64
+	// The list of possible flags are specified by nftnl_rule_attr, see
+	// https://git.netfilter.org/libnftnl/tree/include/libnftnl/rule.h#n21
+	// Current nftables go implementation supports only
+	// NFTNL_RULE_POSITION flag for setting rule at position 0
+	Flags    uint32
 	Exprs    []expr.Any
 	UserData []byte
 }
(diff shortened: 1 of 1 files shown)


rule.go
```diff
@@ -42,6 +42,7 @@ type Rule struct {
 	Chain    *Chain
 	Position uint64
 	Handle   uint64
+	Flags    uint32
```
stapelberg, line 45: Can you add a comment to this field describing which flags are valid to specificy here or which ones the caller should expect? AFAICT, thus far only the NFTA_RULE_POSITION flag plays any role, right?

stapelberg (changes requested):
Thanks for your PR!

stapelberg:
Also, can you add “fixes https://github.com/google/nftables/issues/126” into your commit message please?

turekt:
Done.

stapelberg (approved):

stapelberg:
Thanks!

Issue #115, opened 2020-09-02
bug: GetRule and log prefix rule is not being parsed


greenpau:
It appears that `log prefix` rules are not bring parsed by `GetRule()`

```
table ip filter { # handle 443
        chain forward { # handle 1
                type filter hook forward priority filter; policy drop;
                oifname "cni-podman0" ip daddr 10.88.0.7 tcp dport 80 counter packets 0 bytes 0 accept # handle 4
                log prefix "ip4 forward drop: " # handle 2
                counter packets 0 bytes 0 drop # handle 3
        }
}
```

The following output represents the dump of the above rules.

**Issue**: Notice the `Exprs: ([]expr.Any) <nil>` as it relates to `log prefix` rule.

```
                oifname "cni-podman0" ip daddr 10.88.0.7 tcp dport 80 counter packets 0 bytes 0 accept # handle 4
```

is:

```
(*nftables.Rule)(0xc000267900)({
 Table: (*nftables.Table)(0xc00034a2e0)({
  Name: (string) (len=6) "filter",
  Use: (uint32) 0,
  Flags: (uint32) 0,
  Family: (nftables.TableFamily) 0
 }),
 Chain: (*nftables.Chain)(0xc00032bf80)({
  Name: (string) (len=7) "forward",
  Table: (*nftables.Table)(<nil>),
  Hooknum: (nftables.ChainHook) 0,
  Priority: (nftables.ChainPriority) 0,
  Type: (nftables.ChainType) "",
  Policy: (*nftables.ChainPolicy)(<nil>)
 }),
 Position: (uint64) 0,
 Handle: (uint64) 4,
 Exprs: ([]expr.Any) (len=10 cap=16) {
  (*expr.Meta)(0xc000307b90)({
   Key: (expr.MetaKey) 7,
   SourceRegister: (bool) false,
   Register: (uint32) 1
  }),
  (*expr.Cmp)(0xc00034a3e0)({
   Op: (expr.CmpOp) 0,
   Register: (uint32) 1,
   Data: ([]uint8) (len=16 cap=16) {
    00000000  63 6e 69 2d 70 6f 64 6d  61 6e 30 00 00 00 00 00  |cni-podman0.....|
   }
  }),
  (*expr.Payload)(0xc00034e060)({
   OperationType: (expr.PayloadOperationType) 0,
   DestRegister: (uint32) 1,
   SourceRegister: (uint32) 0,
   Base: (expr.PayloadBase) 1,
[...]

PR #146, opened 2022-02-15
Log prefix expression support

turekt:
Hi,

I have added the `expr.Log` case to the `name` switch of the `unix.NFTA_DATA_EXPR` case as proposed in https://github.com/google/nftables/issues/115 which adds the support for log prefix parsing. I have also enriched your existing `GetRule` test to check the log prefix expression support and whether the addition works correctly.

Let me know what you think.

### DIFF
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -569,6 +569,73 @@ func TestConfigureNATSourceAddress(t *testing.T) {
 	}
 }
 
+func TestExprLogPrefix(t *testing.T) {
+	c, newNS := openSystemNFTConn(t)
+	defer cleanupSystemNFTConn(t, newNS)
+
+	c.FlushRuleset()
+	defer c.FlushRuleset()
+
+	filter := c.AddTable(&nftables.Table{
+		Family: nftables.TableFamilyIPv4,
+		Name:   "filter",
+	})
+	input := c.AddChain(&nftables.Chain{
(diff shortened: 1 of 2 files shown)


stapelberg (changes requested):
Thanks for adding this!

I think it’d be slightly cleaner to not extend the existing TestGetRule, but rather add a new test specifically for log expressions, just like TestDynset.

turekt:
I have extracted the test for log expressions by writing a new test just like the `TestDynset`. I am sure that you are aware, but I want to note that this test has to run with the `-run_system_tests` flag to not be skipped:
```
$ sed -n 256,257p rule.go
					//case "log":
					//	e = &expr.Log{}
$ go test -run_system_tests -run "TestExprLogPrefix"
--- FAIL: TestExprLogPrefix (0.00s)
    nftables_test.go:623: unexpected number of exprs: got 0, want 1
FAIL
exit status 1
FAIL	github.com/google/nftables	0.003s
$ vim rule.go
$ sed -n 256,257p rule.go
					case "log":
						e = &expr.Log{}
$ go test -run_system_tests -run "TestExprLogPrefix"
PASS
ok  	github.com/google/nftables	0.003s
```

Let me know if this is ok.

stapelberg (approved):

stapelberg:
Thanks!

PR #149, opened 2022-03-28
Added support for quota expression

gustavo-iniguez-goya:
Hi!

This PR adds support for quotas.

### DIFF
--- a/expr/quota.go
+++ b/expr/quota.go
@@ -24,15 +24,15 @@ import (
 
 // Quota defines a threshold against a number of bytes.
 type Quota struct {
-	Bytes uint64
-	Used  uint64
-	Over  bool
+	Bytes    uint64
+	Consumed uint64
+	Over     bool
 }
 
 func (q *Quota) marshal() ([]byte, error) {
 	attrs := []netlink.Attribute{
 		{Type: unix.NFTA_QUOTA_BYTES, Data: binaryutil.BigEndian.PutUint64(q.Bytes)},
-		{Type: unix.NFTA_QUOTA_CONSUMED, Data: binaryutil.BigEndian.PutUint64(q.Used)},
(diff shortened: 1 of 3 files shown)


stapelberg:
Thanks for your PR! Could you add a corresponding test function to `nftables_test.go` as well please?

gustavo-iniguez-goya:
Added! sorry, I had some problems figuring out how to extract the want-ed bytes for the tests.
While debugging it I've realized that the flags attribute must be added unconditionally so I've updated quota.go accordingly.

expr/quota.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/google/nftables/binaryutil"
+	"github.com/mdlayher/netlink"
+	"golang.org/x/sys/unix"
+)
+
+// Quota defines a threshold against a number of bytes.
+type Quota struct {
+	Bytes uint64
+	Used  uint64
```
stapelberg, line 28: Should this field be named `Consumed` instead? That would match terminology on the kernel side at least.

nftables_test.go
```diff
@@ -4768,6 +4768,77 @@ func TestNotrack(t *testing.T) {
...
+		Table: filter,
+	})
+
+	c.AddRule(&nftables.Rule{
+		Table: filter,
+		Chain: output,
+		Exprs: []expr.Any{
+			// [ quota bytes 6 consumed 0 flags 1 ]
+			&expr.Quota{
+				Bytes: uint64(6),
```
stapelberg, line 4830: Here and in the next line, I don’t think you need the explicit `uint64(6)`, and just using `6` should work, no?

stapelberg (changes requested):

expr/quota.go
gustavo-iniguez-goya: yes, you're right. I took "used" from the output of `list ruleset`.

stapelberg (approved):

stapelberg:
Thanks!

Issue #151, opened 2022-04-04
Exthdr expression not decoded


Bun:
The following expr.Exthdr (used to clamp TCP MSS) is not decoded, even though the rule was produced by this library. That is, the expr.Exthdr is not present  in the rule.Exprs at all. This rule is equivalent to `nft add rule ip repro repro-chain ip saddr 10.1.0.0/16 tcp flags syn tcp option maxseg size set 1400`.

```
package main

import (
        "fmt"

        "github.com/google/nftables"
        "github.com/google/nftables/expr"
)

var (
        Repro = &nftables.Table{
                Name:   "repro",
                Family: nftables.TableFamilyIPv4,
        }
        ReproChain = &nftables.Chain{
                Name:  "repro-chain",
                Table: Repro,
        }
)

func clampMSS() error {
        var c nftables.Conn

        c.AddTable(Repro)
        if err := c.Flush(); err != nil {
                return err
        }

        c.FlushTable(Repro)
        c.AddChain(ReproChain)
        if err := c.Flush(); err != nil {
                return err
        }

        c.AddRule(&nftables.Rule{
                Table: Repro,
                Chain: ReproChain,
                Exprs: []expr.Any{
                        &expr.Payload{OperationType: 0, DestRegister: 1, Base: expr.PayloadBaseNetworkHeader, Offset: 0xc, Len: 0x4},
                        &expr.Bitwise{SourceRegister: 1, DestRegister: 1, Len: 0x4, Mask: []uint8{255, 255, 0, 0}, Xor: []uint8{0, 0, 0, 0}},
                        &expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: []uint8{10, 1, 0, 0}},

                        // tcp
                        &expr.Meta{Key: expr.MetaKeyL4PROTO, Register: 1},
                        &expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: []uint8{6}},

                        // syn
                        &expr.Payload{DestRegister: 1, Base: expr.PayloadBaseTransportHeader, Offset: 13, Len: 1},
                        &expr.Bitwise{SourceRegister: 1, DestRegister: 1, Len: 0x4, Mask: []uint8{2, 0, 0, 0}, Xor: []uint8{0, 0, 0, 0}},
                        &expr.Cmp{Op: expr.CmpOpNeq, Register: 1, Data: []uint8{0}},

                        // maxseg
[...]

PR #152, opened 2022-04-05
Unmarshal Exthdr and support DestRegister/Flags for reads (#151)

Bun:
For issue #151.

Some fields in Exthdr are context-sensitive and imply a read or write operation. Mixing attributes in an unexpected way will result in EOPNOTSUPP. The exact behavior can be pieced together in the kernel code in `net/netfilter/nft_exthdr.c`

Some example rules to test with:
```
# Read a TCP option
nft add rule ip table forward tcp option maxseg size 1600 drop
# Write a TCP option
nft add rule ip table forward tcp option maxseg size set 1360
```


Bun:
I missed that one. The attributes are now written in the same order that nft writes them in.

stapelberg (approved):

PR #153, opened 2022-04-05
Fix expression parsing data

methril:
Have found different errors on unmarshalling the data expressions for `Range`, `Payload` and `Notrack`:
 * Fix `Range` fields `ToData` and `FromData` unmarshalling.
 * Update the `Payload` field `OperationType` when receiving a `SREG` (follows the same principle that is done while Marshalling)
 * Fix the `Notrack` expression, that has no data associated.


methril:
@stapelberg Could you please take a look to this PR?

It fixes some parsing data when reading the rules with `GetRules`.

stapelberg:
Ah, I didn’t see you signed the CLA in the meantime. Will take a look.

stapelberg:
Thanks!

Issue #154, opened 2022-04-11
GetSetElements - Incorrect port range/interval parsing (or not supported at all)


msdean:
I'm using latest nftables:
nftables (bcad476126f07691b7bc3657d1ade257d469e715)
libmnl (2a3e927ec79e3a52d8cea60ccdfb8699b8a0d56e)
libnftnl (e2514c0eff4da7e8e0aabd410f7b7d0b7564c880)
On kernel 5.13 (Ubuntu 20.04LTS).

I have the following concatenated set:
```
set ip6_set{
                type ipv6_addr . inet_service . ipv6_addr
                flags interval
                elements = {
                             777c:ab4b:85f0:1614:49e5:d29b:aa7b:cc90 . 1-60000 . 8709:1cb9:163e:9b55:357f:ef64:708a:edcb }
        }
```

Emphasis on the port range 1-60000.
When I'm printing the set and its elements I get the following:
The set itself:
```
&nftables.Set{Table:(*nftables.Table)(0xc0000a0160), ID:0x0, Name:"ip6_set", Anonymous:false, Constant:false, Interval:true, IsMap:false, HasTimeout:false, Timeout:0, KeyType:nftables.SetDatatype{Name:"", Bytes:0x0, nftMagic:0x8348}, DataType:nftables.SetDatatype{Name:"", Bytes:0x0, nftMagic:0x0}}
```

The set element:
```
nftables.SetElement{Key:[]uint8{0x77, 0x7c, 0xab, 0x4b, 0x85, 0xf0, 0x16, 0x14, 0x49, 0xe5, 0xd2, 0x9b, 0xaa, 0x7b, 0xcc, 0x90, 0x0, 0x1, 0x0, 0x0, 0x87, 0x9, 0x1c, 0xb9, 0x16, 0x3e, 0x9b, 0x55, 0x35, 0x7f, 0xef, 0x64, 0x70, 0x8a, 0xed, 0xcb}, Val:[]uint8(nil), IntervalEnd:false, VerdictData:(*expr.Verdict)(nil), Timeout:0}
```

Specifically, note the middle part that corresponds to the port range (after the `0x90` and before the `0x87`):
```
nftables.SetElement{Key:[]uint8{...0x90, 0x0, 0x1, 0x0, 0x0, 0x87,...}
```

As you can see, the first value is parsed correctly (0x0, 0x1 == 1), but the other part isn't (0x0, 0x0 == 0).

[...]

PR #155, opened 2022-04-20
Support for concat set intervals

turekt:
Hi,

it seems that the current implementation of the nftables go lib does not support what was mentioned in issue #154. 

The root cause for the interval decode failure is that the lib does not handle the `NFTA_SET_ELEM_KEY_END` flag in the [`decode` function](https://github.com/google/nftables/blob/16a134723a9679a5cc4a63fe37f7624ccf7c07f9/set.go#L247-L263). Since `NFTA_SET_ELEM_KEY_END` is not a recognized flag, marshaling of intervals in a concatenated type is not supported as well.

This pull request handles intervals in concatenations and contains tests for newly introduced code. Some changes were made to existing code to fully support this feature. I was not able to find some of the constants in the unix go package (see code comments), therefore I have added them as set constants as a preliminary approach at least until the missing constants are added to the unix go package.

Let me know what you think.


set_test.go
```diff
@@ -137,3 +137,32 @@ func TestConcatSetType(t *testing.T) {
 		})
 	}
 }
+
+func TestConcatSetTypeElements(t *testing.T) {
+	t.Parallel()
+
+	tests := map[string][]SetDatatype{
```
stapelberg, line 144: Can you make this a slice of struct please, so that the iteration order (and hence test run order) is deterministic?

stapelberg (approved):

set_test.go
turekt: Good point.

stapelberg (approved):

stapelberg:
Thanks!

Issue #157, opened 2022-05-02
`Conn` isn't a connection, but a network namespace reference; could it please getting refactored into connection with Open/Dial and Close?


thediveo:
At this time, `Conn` is something resembling more of an address, or network namespace address (reference) to be more precise. Additionally, it has an exported(!) mutex factored in which is kind of strange it it should be unexported?

```go
type Conn struct {
	TestDial nltest.Func // for testing only; passed to nltest.Dial
	NetNS    int         // Network namespace netlink will interact with.
	sync.Mutex
	// contains filtered or unexported fields
}
```

This current design assumedly comes from its current most prominent use case: set up nftable changes and then apply (flush) them. In consequence, the connection-disguised netns address is only opened temporarily for each individual operation on nftables, such as a flush of pending changing.

In consequence, discovering the current state of nftables is highly inefficient, as it requires a large amount of NETLINK socket opening and closing in rapid repeated sequence. To properly discover the current nftables setup, such as for running diagnosis tools on it, the following steps are required, resulting in constant socket opening and closing for each individual (repeated) step:

1. `conn.ListTables()` -- in order to get table properties not returned implicitly by `conn.ListChains()`
2. `conn.ListChains()`
3. for each chain:
   1. `conn.GetRule()` -- misnomer, it actually works like `GetRules()`

My humble suggestion would be:
- refactor `Conn` by adding `Open` and `Close()` receivers to allow for improved behavior. At least for some time these can be co-exist with the current implicit behavior when calling nftables operations like `ListTables()` without `Open` first.
  - `Open(opts)` could make use of the option (function) pattern to allow to custom dialing or even passing in a suitable fd opened by some other means, this would care for issue #132 (which I would like to see covered not least as I have my own namespace discovery engine module [lxkns](https://github.com/thediveo/lxkns) also covering switching namespaces and covering for many more cases than the well-known netns package).
  - `Close()` would be allowed to be called multiple times, but subject to the now-internal connection mutex as to not close the NETLINK socket while in flight.
- make `sync.Mutex` non-exported -- unless I missed a reason for exporting it?
- add `GetRules()` receiver and deprecate `GetRule()` for slow migration.

I'm willing to work on a PR given the signal that there is actual interest in guiding me through my foolish mistakes and finally integrating it.

Any things you can suggest already upfront?

stapelberg:
[...]

PR #159, opened 2022-05-07
implement "lasting" connections, issue #157

thediveo:
- add New constructor, Close receiver to optionally support "lasting" netlink connections while defaulting to existing temporary netlink connection usage
- add unit test for New, Close and correct default connection handling behavior
- make Conn mutex un-exported
- refactor tests to use New constructor
- ~~rename GetRule to GetRules and deprecate GetRule, update usage~~ broken out into separate PR #160

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -32,18 +32,138 @@ import (
 // Commands are buffered. Flush sends all buffered commands in a single batch.
 type Conn struct {
 	TestDial nltest.Func // for testing only; passed to nltest.Dial
-	NetNS    int         // Network namespace netlink will interact with.
-	sync.Mutex
+	NetNS    int         // fd referencing the network namespace netlink will interact with.
+
+	lasting  bool       // establish a lasting connection to be used across multiple netlink operations.
+	mu       sync.Mutex // protects the following state
 	messages []netlink.Message
 	err      error
+	nlconn   *netlink.Conn // netlink socket using NETLINK_NETFILTER protocol.
+}
+
+// ConnOption is an option to change the behavior of the nftables Conn returned by Open.
+type ConnOption func(*Conn)
+
+// New returns a netlink connection for querying and modifying nftables. Some
+// aspects of the new netlink connection can be configured using the options
+// WithNetNSFd, WithTestDial, and AsLasting.
+//
+// A lasting netlink connection should be closed by calling CloseLasting() to
+// close the underlying lasting netlink connection, cancelling all pending
+// operations using this connection.
+func New(opts ...ConnOption) (*Conn, error) {
+	cc := &Conn{}
+	for _, opt := range opts {
+		opt(cc)
+	}
+
+	if !cc.lasting {
+		return cc, nil
+	}
+
+	nlconn, err := cc.dialNetlink()
+	if err != nil {
+		return nil, err
+	}
+	cc.nlconn = nlconn
+	return cc, nil
+}
+
+// AsLasting creates the new netlink connection as a lasting connection that is
+// reused across multiple netlink operations, instead of opening and closing the
+// underlying netlink connection only for the duration of a single netlink
+// operation.
+func AsLasting() ConnOption {
+	return func(cc *Conn) {
+		// We cannot create the underlying connection yet, as we are called
+		// anywhere in the option processing chain and there might be later
+		// options still modifying connection behavior.
+		cc.lasting = true
+	}
+}
+
+// WithNetNSFd sets the network namespace to create a new netlink connection to:
+// the fd must reference a network namespace.
+func WithNetNSFd(fd int) ConnOption {
+	return func(cc *Conn) {
+		cc.NetNS = fd
+	}
+}
+
+// WithTestDial sets the specified nltest.Func when creating a new netlink
+// connection.
+func WithTestDial(f nltest.Func) ConnOption {
+	return func(cc *Conn) {
+		cc.TestDial = f
+	}
+}
+
+// netlinkCloser is returned by netlinkConn(UnderLock) and must be called after
+// being done with the returned netlink connection in order to properly close
+// this connection, if necessary.
+type netlinkCloser func() error
+
+// netlinkConn returns a netlink connection together with a netlinkCloser that
+// later must be called by the caller when it doesn't need the returned netlink
+// connection anymore. The netlinkCloser will close the netlink connection when
+// necessary. If New has been told to create a lasting connection, then this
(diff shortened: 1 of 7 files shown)


conn.go
```diff
@@ -32,18 +32,125 @@ import (
...
+	// instead only to make sure to only run the netlink socket close on the
+	// first time with an open long-running netlink socket. As Open() acts as
+	// constructor, it is not possible to re-Open an already closed connection.
+	cc.mu.Lock()
+	nlconn := cc.nlconn
+	cc.nlconn = nil
+	cc.mu.Unlock()
+	if nlconn != nil {
+		var err error
+		cc.closeOnce.Do(func() {
```
stapelberg, line 140: Isn’t closeOnce superfluous when you’re already using a mutex?

conn.go
```diff
@@ -32,18 +32,125 @@ import (
...
+}
+
+// ConnOption is an option to change the behavior of the nftables Conn returned by Open.
+type ConnOption func(*Conn)
+
+// Open returns a ("long-running") netlink connection for querying and modifying
+// nftables. It should be closed by calling Close() to close the underlying
+// netlink connection and release associated resources and pending operations.
+//
+// Please note that it is not necessary to call Open in order to create an
```
stapelberg, line 51: I think once a constructor exists, people will be drawn to using it, even if not strictly necessary.

Instead of documenting that a Conn{} can be used, how about we make the long-running connection bit a ConnOption, and have Open() behave the same way as Conn{} now

stapelberg (changes requested):
Thanks for sending the PR! Could you split the the mutex and GetRule changes into separate pull requests (or at least separate commits) please?

conn.go
thediveo: yes, agreed ... but wasn't exactly sure to not overlook some corner case.

conn.go
thediveo: should we then still name it `Open` or rather `New?` And then an `KeepOpen` option?

stapelberg: I think `New` would be a slightly better name, yeah.

conn.go
thediveo: removed superfluous closeOnce

conn.go
thediveo: refactored accordingly with `New` and `AsLasting` for requesting a "lasting" connection. Also refactored tests to use the constructor instead of creating `Conn` instances directly.

conn.go
```diff
@@ -32,18 +32,131 @@ import (
...
+// AsLasting option when ncreating this connection. If either no lasting netlink
+// connection has been opened or the (not-so) lasting connection is already in
+// the process of closing or has been closed, Close will immediately return
+// without any error.
+func (cc *Conn) Close() error {
+	// Don't acquire the lock for the whole duration of the Close operation, but
+	// instead only so long as to make sure to only run the netlink socket close
+	// on the first time with a lasting netlink socket. As there is only the
+	// New() constructor, but no Open() method, it's impossible to reopen a
+	// lasten connection.
```
stapelberg, line 143: typo: lasten → lasting

conn.go
```diff
@@ -32,18 +32,131 @@ import (
...
+		return nil, nil, err
+	}
+	return nlconn, func() error { return nlconn.Close() }, nil
+}
+
+// Close closes the lasting netlink connection that has been opened using
+// AsLasting option when ncreating this connection. If either no lasting netlink
+// connection has been opened or the (not-so) lasting connection is already in
+// the process of closing or has been closed, Close will immediately return
+// without any error.
```
stapelberg, line 137: Can you add a paragraph that explains whether it’s safe to keep using a Conn after calling Close() — will it fall back from a lasting connection into an on-demand connection gracefully?

conn.go
```diff
@@ -32,18 +32,131 @@ import (
...
+// ConnOption is an option to change the behavior of the nftables Conn returned by Open.
+type ConnOption func(*Conn)
+
+// New returns a netlink connection for querying and modifying nftables. Some
+// aspects of the new netlink connection can be configured using the options
+// WithNetNSFd, WithTestDial, and AsLasting.
+//
+// A lasting netlink connection should be closed by calling Close() to close the
+// underlying netlink connection and release the associated resources and
+// pending operations.
```
stapelberg, line 53: Is “release the […] pending operations” really the best verb? It will cancel any pending operations, right? Maybe we should use “cancel” as verb to clarify this.

stapelberg (commented):
Thanks, this is looking good overall! Couple of small comments. Can you rebase on current master branch contents, too, please?

conn.go
```diff
@@ -32,18 +32,138 @@ import (
...
+	nlconn, err := cc.dialNetlink()
+	if err != nil {
+		return nil, nil, err
+	}
+	return nlconn, func() error { return nlconn.Close() }, nil
+}
+
+// Close closes the lasting netlink connection that has been opened using
+// AsLasting option when creating this connection. If either no lasting netlink
+// connection has been opened or the (not-so) lasting connection is already in
```
stapelberg, line 135: > or the (not-so) lasting connection is already in the process of closing or has been closed

Is this correct? AFAICT, the function does not work with the non-lasting connections that `netlinkConnUnderLock()` returns, because they are never assigned to `cc.nlconn`, no?

conn.go
```diff
@@ -32,18 +32,138 @@ import (
...
+// the process of closing or has been closed, Close will immediately return
+// without any error.
+//
+// Close will terminate all pending netlink operations using the lasting
+// connection.
+//
+// After closing a lasting connection, the connection will revert to using
+// on-demand transient netlink connections when calling further netlink
+// operations (such as GetTables).
+func (cc *Conn) Close() error {
```
stapelberg, line 145: Given that `Close` only has an effect on the lasting connection, can we rename it to `CloseLasting`?

conn.go
thediveo: I suspect that my twist-in-tongue meant "(not-so) lasting connection" sent you in the wrong direction -- and as documentation should be absolutely clear in such places I've simply removed the "(not-so)" in order to make clear that this applies to the lasting connection, not the on-demand connections. The sentence then should hold as `Close` immediately returns in both described cases. I hope that this makes things clear?

conn.go
thediveo: While this is an unusual name I see the reason in this particular case, because we have the unusual situation of both on-demand as well as lasting connections. I've refactored the name.

stapelberg (approved):

stapelberg:
Thank you very much!

PR #160, opened 2022-05-08
add GetRules and deprecate GetRule, update tests

thediveo:
Breaks out the `GetRule` to `GetRules` refactoring from issue #157  into a separate PR, as requested in PR #159.


stapelberg (approved):

stapelberg:
Thanks!

Issue #161, opened 2022-05-10
Unsupported (data) expressions "target" and "match"


thediveo:
When getting the rules of all chains of all tables in the host of an Ubuntu 21.10 system I notice missing expressions. For testing, I've made the currently "silent" `exprsFromMsg` function to dump the names of currently unimplemented `NFTA_EXPR_DATA` names. This shows `target` and `match` data expressions.

Would you like a PR for these two types?

A few questions about un/marshalling strings which are per netfilter source `char[fixelen]`, such as the `name` attribute in a  target data structure:
1. should the Go type stick with `string` instead of a fixed-size array?
2. would unmarshalling `string(bytes.Trim(ad.Bytes(), "\x00"))` be the correct way instead of `ad.String()`?

stapelberg:
> Would you like a PR for these two types?

Sure!

> A few questions about un/marshalling strings which are per netfilter source `char[fixelen]`, such as the `name` attribute in a target data structure:

Can you send a link to the definition please?

> * should the Go type stick with `string` instead of a fixed-size array?

Probably string + length check would be the best way. That way, should the fixed size array ever grow larger in the kernel, we can easily adapt without breaking anyone.

> * would unmarshalling `string(bytes.Trim(ad.Bytes(), "\x00"))` be the correct way instead of `ad.String()`?

expr/verdict.go uses a similar approach: https://github.com/google/nftables/blob/eeaebcf552951a01af8c359bec279367468aca28/expr/verdict.go#L118

thediveo:
> > A few questions about un/marshalling strings which are per netfilter source `char[fixelen]`, such as the `name` attribute in a target data structure:
> 
> Can you send a link to the definition please?

I've included the asked information in PR #163, so we can discuss on the code itself.

thediveo:
Done.

PR #163, opened 2022-05-11
un/marshal Match and Target expressions

thediveo:
- implements issue #161
- implements marshaling and un-marshaling `Match` and `Target` expressions.
- implements unit tests for `Match` and `Target`.

Both `Match` and `Target` have the same structure, where only the netlink tag names differ; `unix.NFTA_MATCH_...` versus `unix.NFTA_TARGET_...`. The following discussion is sufficient to address only `Match`, with `Target` being basically the same.

>> A few questions about un/marshalling strings which are per netfilter source char[fixelen], such as the name attribute in a target data structure:
>
> Can you send a link to the definition please?

- [`nftl_expr_match`](https://git.netfilter.org/libnftnl/tree/src/expr/match.c?id=09456c720e9c00eecc08e41ac6b7c291b3821ee5#n30) is `libnftl`'s view;  to be honest, the way `libnftl` deals with the border cases of a fixed length array without a string terminating \0 in those cases -- it doesn't -- looks slightly sketchy to me.
   ```c
  struct nftnl_expr_match {
	  char		name[XT_EXTENSION_MAXNAMELEN];
	  uint32_t	rev;
	  uint32_t	data_len;
	  const void	*data;
  };
   ```
- the Linux kernel side in [`net/netfilter/nft_compat.c`](https://elixir.bootlin.com/linux/latest/source/net/netfilter/nft_compat.c#L413) looks as follows:
  ```c
   static const struct nla_policy nft_match_policy[NFTA_MATCH_MAX + 1] = {
       [NFTA_MATCH_NAME]	= { .type = NLA_NUL_STRING },
       [NFTA_MATCH_REV]	= { .type = NLA_U32 },
       [NFTA_MATCH_INFO]	= { .type = NLA_BINARY },
   };
   ```

Would you agree that the correct marshaling and un-marshaling should use an "unlimited" string representation with trailing \0 and not care about the libnftl internal fixed length storage?

### DIFF
--- a/expr/match.go
+++ b/expr/match.go
@@ -1,0 +1,63 @@
+package expr
+
+import (
+	"bytes"
+	"encoding/binary"
+
+	"github.com/google/nftables/binaryutil"
+	"github.com/mdlayher/netlink"
+	"golang.org/x/sys/unix"
+)
+
+// See https://git.netfilter.org/libnftnl/tree/src/expr/match.c?id=09456c720e9c00eecc08e41ac6b7c291b3821ee5#n30
+type Match struct {
+	Name string
+	Rev  uint32
+	Info []byte
+}
+
+func (e *Match) marshal() ([]byte, error) {
+	// Per https://git.netfilter.org/libnftnl/tree/src/expr/match.c?id=09456c720e9c00eecc08e41ac6b7c291b3821ee5#n38
+	name := e.Name
+	// limit the extension name as (some) user-space tools do and leave room for
+	// trailing \x00
+	if len(name) >= /* sic! */ XTablesExtensionNameMaxLen {
+		name = name[:XTablesExtensionNameMaxLen-1] // leave room for trailing \x00.
+	}
+	attrs := []netlink.Attribute{
+		{Type: unix.NFTA_MATCH_NAME, Data: []byte(name + "\x00")},
+		{Type: unix.NFTA_MATCH_REV, Data: binaryutil.BigEndian.PutUint32(e.Rev)},
+		{Type: unix.NFTA_MATCH_INFO, Data: e.Info},
+	}
+	data, err := netlink.MarshalAttributes(attrs)
+	if err != nil {
+		return nil, err
+	}
+
+	return netlink.MarshalAttributes([]netlink.Attribute{
+		{Type: unix.NFTA_EXPR_NAME, Data: []byte("match\x00")},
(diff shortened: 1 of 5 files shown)


expr/match.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return err
+	}
+
+	ad.ByteOrder = binary.BigEndian
+	for ad.Next() {
+		switch ad.Type() {
+		case unix.NFTA_MATCH_NAME:
+			// We are forgiving here, accepting any length and missing terminating \x00.
+			e.Name = string(bytes.Trim(ad.Bytes(), "\x00"))
```
stapelberg, line 54: Here and elsewhere, can you use `bytes.TrimRight`? Shouldn’t matter in practice, but just to be as precise as possible :)

expr/match.go
```diff
@@ -0,0 +1,62 @@
...
+type Match struct {
+	Name string
+	Rev  uint32
+	Info []byte
+}
+
+func (e *Match) marshal() ([]byte, error) {
+	// Per https://git.netfilter.org/libnftnl/tree/src/expr/match.c?id=09456c720e9c00eecc08e41ac6b7c291b3821ee5#n38
+	name := e.Name
+	// leave room for trailing \x00; this looks like a weak spot in libnftnl
```
stapelberg, line 22: Can you elaborate on what weak spot you see?

AFAICT, `libnftnl` uses:
```c
		snprintf(mt->name, sizeof(mt->name), "%.*s", data_len,
			 (const char *)data);
```

And `snprintf` is documented to write at most `sizeof(mt->name)` bytes, including the terminating null byte.

So the size limit here is 29 bytes, including the terminating null byte.

But the Go code you have permits 29 bytes *plus* an extra null byte.

stapelberg (changes requested):
> Would you agree that the correct marshaling and un-marshaling should use an "unlimited" string representation with trailing \0 and not care about the libnftl internal fixed length storage?

We should be permissive in what we accept (unmarshal into unlimited string), but strict in what we generate (uphold the existing size limit, even if it’s only for compatibility with other userspace tools).

expr/match.go
thediveo: Ah, I missed the part that `snprintf` ensures to correctly terminate with `\x00`. So my comment is incorrect and I'll remove it; instead, I will "trim" this to 29 octets 😁 plus terminating zero octet.

stapelberg (approved):

stapelberg:
Thanks!

PR #164, opened 2022-05-15
typed Info field payloads for Match and Target expressions 

thediveo:
In order to implement some first xtables types to be used with `Match.Info` and `Target.Info` fields I needed to prepare some infrastructure first. This PR is thus roughly divided into parts three.

1. (un)marshalling xtable-related information unfortunately(?) needs in several cases the table family, such as when dealing with IPv4/IPv6 addresses in their original C structures with host endianess. The first commit in this series refactors the existing expression un/marshal functions (`expr.Any`, ...) to now properly pass this information down to the place where the `Match.Info` and `Target.Info` fields will require this information. For this to work, a very few places in the `nftables` package need to be updated in order to pass the required table family information. No existing `nftables` API need to be changed for this. The existing unit tests are all passing as before.

2. xtables information in `Match.Info` and `Target.Info` unfortunately is in host endianess and alignment. This prohibits us from using the existing netlink TLV de- and encoders. This commit adds a new `alignedbuffer` package (incl. unit tests) to un/marshal xtables information, paying particular attention to some slightly untypical details, such as padding xtables information at its end to the largest (uint64) host alignment, 32bit aligned byte sequences, and some more. Unit tests are included.

3. the final commit adds un/marshalling of several xtables-related types for the matchers/targets addrtype, conntrack (3 different revisions!), udp, tcp, dnat, masquerade. This commit covers the iptables chains/rules to be properly handled as used by Docker CE 21.x. Unit tests for round-trip marshalling-unmarshalling are included.

### DIFF
--- a/alignedbuff/alignedbuff.go
+++ b/alignedbuff/alignedbuff.go
@@ -1,0 +1,240 @@
+// Package alignedbuff implements encoding and decoding aligned data elements
+// to/from buffers in native endianess.
+package alignedbuff
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"unsafe"
+
+	"github.com/google/nftables/binaryutil"
+)
+
+// ErrEOF signals trying to read beyond the available payload information.
+var ErrEOF = errors.New("not enough data left")
+
+// AlignedBuff implements marshalling and unmarshalling information in
+// platform/architecture native endianess and data type alignment. It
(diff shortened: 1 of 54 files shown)


stapelberg (approved):

stapelberg:
What an imprresive amount of work! Thanks!

PR #165, opened 2022-05-15
use TableFamilyUnspecified (NFPROTO_UNSPEC) instead of AF_UNSPEC

thediveo:
The `NFPROTO_` constant definitions for the various netfilter table families also feature `NFPROTO_UNSPEC`. While its value match the value of `AF_UNSPEC`, for clarity the (currently missing) `TableFamilyUnspecified` should be used in `nftables` code in the context of `extraHeader` instead of `AF_UNSPEC`.


stapelberg:
Thanks!

PR #168, opened 2022-06-03
list tables and chains optionally by specific table family

thediveo:
- adds new ListChainsOfTableFamily() to supplement existing ListChains()
- adds new ListTablesOfFamily() to supplement existing ListTables()

### DIFF
--- a/table.go
+++ b/table.go
@@ -101,6 +101,12 @@ func (cc *Conn) FlushTable(t *Table) {
 
 // ListTables returns currently configured tables in the kernel
 func (cc *Conn) ListTables() ([]*Table, error) {
+	return cc.ListTablesOfFamily(TableFamilyUnspecified)
+}
+
+// ListTables returns currently configured tables for the specified table family
+// in the kernel. It lists all tables if family is TableFamilyUnspecified.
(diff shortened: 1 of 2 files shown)


table.go
```diff
@@ -101,6 +101,12 @@ func (cc *Conn) FlushTable(t *Table) {
 
 // ListTables returns currently configured tables in the kernel
 func (cc *Conn) ListTables() ([]*Table, error) {
+	return cc.ListTablesOfFamily(TableFamilyUnspecified)
+}
+
+// ListTables returns currently configured tables for the specified table family
+// in the kernel. It lists all tables if family is TableFamilyUnspecified.
+func (cc *Conn) ListTablesOfFamily(family TableFamily) ([]*Table, error) {
```
stapelberg, line 109: You’re passing `family`, but you’re not using it yet in line 121 (which still reads TableFamilyUnspecified).

stapelberg (changes requested):

table.go
thediveo: D'oh! Fixed.

stapelberg (approved):

Issue #172, opened 2022-07-08
Is dynamic set or meters supported?


zjuchenchao:
I find keyword meters [here](https://wiki.nftables.org/wiki-nftables/index.php/Meters), and I'd like to limit rate per second like this

```
nft add rule ip filter INPUT meter myMeter { ip saddr limit rate 200/second } counter drop --debug=netlink
myMeter filter 21
myMeter filter 0
ip filter INPUT 
  [ payload load 4b @ network header + 12 => reg 1 ]
  [ dynset add reg_key 1 set myMeter expr [ limit rate 200/second burst 0 type packets flags 0x0 ] ]
  [ counter pkts 0 bytes 0 ]
  [ immediate reg 0 drop ]
```
The 2nd expr should be a dynset, but how can I add the _expr [ limit rate 200/second burst 0 type packets flags 0x0 ]_ to the dynset?
https://pkg.go.dev/github.com/google/nftables#Set
https://pkg.go.dev/github.com/google/nftables/expr#Dynset
https://pkg.go.dev/github.com/google/nftables/expr#Limit
I looked it up on go.dev. Maybe I should add the limit expr to a set, and add SetID and SetName to the dynset, but SetDatatype cannot be a expr. Is there a way to make it or am I thinking wrong? Thanks a lot.

stapelberg:
Our expr.Dynset implementation does not currently support expressions. See the upstream libnftnl code for confirmation: https://sources.debian.org/src/libnftnl/1.2.2-1/src/expr/dynset.c/?hl=167#L167 — that code marshals `NFTA_DYNSET_EXPR` (single expression) or `NFTA_DYNSET_EXPRESSIONS` (multiple expressions) in addition to the attributes that we already have.

Feel free to send a PR to add expression support to expr.Dynset.

Furisto:
What about support for meters? I am trying to implement this
`nft add rule ip filter output ip protocol tcp ct state new meter ratemeter '{ ip daddr & 0.0.0.0 timeout 1m limit rate over 10/minute }' counter drop`

but do  not see any way to express that. Any pointers?

stapelberg:
> nft add rule ip filter output ip protocol tcp ct state new meter ratemeter '{ ip daddr & 0.0.0.0 timeout 1m limit rate over 10/minute }' counter drop

By adding `--debug=all`, we can see:

```
[...]

PR #173, opened 2022-07-25
Added dynset expressions support

turekt:
Hi,

I have added support for dynset expressions as suggested in #172. The pull request contains a bit more than just adding support for dynset expressions:
- Function `exprFromMsg` is now moved to the `expr` package and made public
- Limit expression marshaling logic is changed to better correspond to nftables messages
- Added separate tests for one and multiple dynset expressions (also demonstrates how to create meters)
- Added additional dynamic flag to sets which was missing in order to fully support meters
- Added a small set of new constants which are still not supported by unix package
- Added an additional connlimit expr implementation which was not supported

Let me know what you think.

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -19,9 +19,40 @@ import (
 	"encoding/binary"
 
 	"github.com/google/nftables/binaryutil"
+	"github.com/google/nftables/internal/parseexprfunc"
 	"github.com/mdlayher/netlink"
 	"golang.org/x/sys/unix"
 )
+
+func init() {
+	parseexprfunc.ParseExprBytesFunc = func(fam byte, ad *netlink.AttributeDecoder, b []byte) ([]interface{}, error) {
+		exprs, err := exprsFromBytes(fam, ad, b)
+		if err != nil {
+			return nil, err
+		}
+		result := make([]interface{}, len(exprs))
+		for idx, expr := range exprs {
+			result[idx] = expr
+		}
+		return result, nil
+	}
+	parseexprfunc.ParseExprMsgFunc = func(fam byte, b []byte) ([]interface{}, error) {
+		ad, err := netlink.NewAttributeDecoder(b)
+		if err != nil {
+			return nil, err
+		}
+		ad.ByteOrder = binary.BigEndian
+		var exprs []interface{}
+		for ad.Next() {
+			e, err := parseexprfunc.ParseExprBytesFunc(fam, ad, b)
+			if err != nil {
+				return e, err
+			}
+			exprs = append(exprs, e...)
+		}
+		return exprs, ad.Err()
+	}
+}
 
 // Marshal serializes the specified expression into a byte slice.
 func Marshal(fam byte, e Any) ([]byte, error) {
@@ -31,6 +62,96 @@ func Marshal(fam byte, e Any) ([]byte, error) {
 // Unmarshal fills an expression from the specified byte slice.
 func Unmarshal(fam byte, data []byte, e Any) error {
 	return e.unmarshal(fam, data)
+}
+
+// exprsFromBytes parses nested raw expressions bytes
+// to construct nftables expressions
+func exprsFromBytes(fam byte, ad *netlink.AttributeDecoder, b []byte) ([]Any, error) {
+	var exprs []Any
+	ad.Do(func(b []byte) error {
+		ad, err := netlink.NewAttributeDecoder(b)
+		if err != nil {
+			return err
+		}
+		ad.ByteOrder = binary.BigEndian
+		var name string
+		for ad.Next() {
+			switch ad.Type() {
+			case unix.NFTA_EXPR_NAME:
+				name = ad.String()
+				if name == "notrack" {
+					e := &Notrack{}
+					exprs = append(exprs, e)
+				}
+			case unix.NFTA_EXPR_DATA:
+				var e Any
+				switch name {
+				case "ct":
+					e = &Ct{}
+				case "range":
+					e = &Range{}
+				case "meta":
+					e = &Meta{}
+				case "cmp":
+					e = &Cmp{}
+				case "counter":
+					e = &Counter{}
+				case "payload":
+					e = &Payload{}
+				case "lookup":
+					e = &Lookup{}
+				case "immediate":
+					e = &Immediate{}
+				case "bitwise":
+					e = &Bitwise{}
+				case "redir":
+					e = &Redir{}
+				case "nat":
+					e = &NAT{}
+				case "limit":
+					e = &Limit{}
+				case "quota":
+					e = &Quota{}
+				case "dynset":
+					e = &Dynset{}
+				case "log":
+					e = &Log{}
+				case "exthdr":
+					e = &Exthdr{}
+				case "match":
+					e = &Match{}
+				case "target":
+					e = &Target{}
+				case "connlimit":
+					e = &Connlimit{}
+				}
+				if e == nil {
+					// TODO: introduce an opaque expression type so that users know
+					// something is here.
+					continue // unsupported expression type
+				}
+
+				ad.Do(func(b []byte) error {
+					if err := Unmarshal(fam, b, e); err != nil {
+						return err
+					}
+					// Verdict expressions are a special-case of immediate expressions, so
+					// if the expression is an immediate writing nothing into the verdict
+					// register (invalid), re-parse it as a verdict expression.
+					if imm, isImmediate := e.(*Immediate); isImmediate && imm.Register == unix.NFT_REG_VERDICT && len(imm.Data) == 0 {
+						e = &Verdict{}
+						if err := Unmarshal(fam, b, e); err != nil {
+							return err
+						}
+					}
+					exprs = append(exprs, e)
+					return nil
+				})
+			}
+		}
+		return ad.Err()
+	})
+	return exprs, ad.Err()
 }
 
 // Any is an interface implemented by any expression type.
--- a/expr/limit.go
+++ b/expr/limit.go
@@ -72,24 +72,16 @@ type Limit struct {
 }
 
 func (l *Limit) marshal(fam byte) ([]byte, error) {
+	var flags uint32
+	if l.Over {
+		flags = unix.NFT_LIMIT_F_INV
+	}
 	attrs := []netlink.Attribute{
-		{Type: unix.NFTA_LIMIT_TYPE, Data: binaryutil.BigEndian.PutUint32(uint32(l.Type))},
 		{Type: unix.NFTA_LIMIT_RATE, Data: binaryutil.BigEndian.PutUint64(l.Rate)},
 		{Type: unix.NFTA_LIMIT_UNIT, Data: binaryutil.BigEndian.PutUint64(uint64(l.Unit))},
-	}
-
-	if l.Over {
-		attrs = append(attrs, netlink.Attribute{
-			Type: unix.NFTA_LIMIT_FLAGS,
-			Data: binaryutil.BigEndian.PutUint32(unix.NFT_LIMIT_F_INV),
-		})
-	}
-
-	if l.Burst != 0 {
-		attrs = append(attrs, netlink.Attribute{
-			Type: unix.NFTA_LIMIT_BURST,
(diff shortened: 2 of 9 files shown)


expr/limit.go
```diff
@@ -72,24 +72,16 @@ type Limit struct {
...
-			Data: binaryutil.BigEndian.PutUint32(unix.NFT_LIMIT_F_INV),
-		})
-	}
-
-	if l.Burst != 0 {
-		attrs = append(attrs, netlink.Attribute{
-			Type: unix.NFTA_LIMIT_BURST,
-			Data: binaryutil.BigEndian.PutUint32(l.Burst),
-		})
+		{Type: unix.NFTA_LIMIT_BURST, Data: binaryutil.BigEndian.PutUint32(l.Burst)},
```
stapelberg, line 82: Why is the l.Burst handling no longer conditional?

expr/expr.go
```diff
@@ -33,6 +33,115 @@ func Unmarshal(fam byte, data []byte, e Any) error {
...
+			}
+		}
+		return ad.Err()
+	})
+	return exprs, ad.Err()
+}
+
+// ExprsFromMsg parses netlink message in order to
+// extract and construct expressions
+func ExprsFromMsg(fam byte, b []byte) ([]Any, error) {
```
stapelberg, line 128: It’s not clear to me why these two functions need to be exported? Can you clarify please?

stapelberg (commented):
Thanks for your PR!

expr/limit.go
turekt: I observed that burst is passed with nft regardless of whether it is specified or not:
```
# nft --debug=netlink add rule ip filter forward add @myMeter { ip saddr ct count over 20 limit rate over 10/second } drop
ip filter forward
  [ payload load 4b @ network header + 12 => reg 1 ]
  [ dynset add reg_key 1 set myMeter expr [ connlimit count 20 flags 1 ] expr [ limit rate 10/second burst 5 type packets flags 0x1 ] ]
  [ immediate reg 0 drop ]
```

The same is observed from the raw bytes when burst is not specified in the nft command:
```
# strace -f -v -x -s 2048 -eraw=sendto nft add rule ip filter forward add @myMeter { ip saddr limit rate over 10/second } drop
...
sendmsg(3, {msg_name={sa_family=AF_NETLINK, nl_pid=0, nl_groups=00000000}, msg_namelen=12, msg_iov=[{iov_base=[[{nlmsg_len=20, nlmsg_type=NFNL_MSG_BATCH_BEGIN, nlmsg_flags=NLM_F_REQUEST, nlmsg_seq=0, nlmsg_pid=0}, {nfgen_family=AF_UNSPEC, version=NFNETLINK_V0, res_id=htons(2560)}], [{nlmsg_len=272, nlmsg_type=NFNL_SUBSYS_NFTABLES<<8|NFT_MSG_NEWRULE, nlmsg_flags=NLM_F_REQUEST|NLM_F_CREATE|NLM_F_APPEND, nlmsg_seq=1, nlmsg_pid=0}, {nfgen_family=AF_INET, version=NFNETLINK_V0, res_id=htons(0)}, [[{nla_len=11, nla_type=0x1}, "\x66\x69\x6c\x74\x65\x72\x00"], [{nla_len=12, nla_type=0x2}, "\x66\x6f\x72\x77\x61\x72\x64\x00"], [{nla_len=228, nla_type=NLA_F_NESTED|0x4}, "\x34\x00\x01\x80\x0c\x00\x01\x00\x70\x61\x79\x6c\x6f\x61\x64\x00\x24\x00\x02\x80\x08\x00\x01\x00\x00\x00\x00\x01\x08\x00\x02\x00\x00\x00\x00\x01\x08\x00\x03\x00\x00\x00\x00\x0c\x08\x00\x04\x00\x00\x00\x00\x04\x7c\x00\x01\x80\x0b\x00\x01\x00\x64\x79\x6e\x73\x65\x74\x00\x00\x6c\x00\x02\x80\x08\x00\x04\x00\x00\x00\x00\x01\x08\x00\x03\x00\x00\x00\x00\x00\x0c\x00\x01\x00\x6d\x79\x4d\x65\x74\x65\x72\x00\x08\x00\x02\x00\x00\x00\x00\x01\x44\x00\x07\x80\x0a\x00\x01\x00\x6c\x69\x6d\x69\x74\x00\x00\x00\x34\x00\x02\x80\x0c\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x0a\x0c\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x01\x08\x00\x03\x00\x00\x00\x00\x05\x08\x00\x04\x00\x00\x00\x00\x00\x08\x00\x05\x00\x00\x00\x00\x01\x30\x00\x01\x80\x0e\x00\x01\x00\x69\x6d\x6d\x65\x64\x69\x61\x74\x65\x00\x00\x00\x1c\x00\x02\x80\x08\x00\x01\x00\x00\x00\x00\x00\x10\x00\x02\x80\x0c\x00\x02\x80\x08\x00\x01\x00\x00\x00\x00\x00"]]], [{nlmsg_len=20, nlmsg_type=NFNL_MSG_BATCH_END, nlmsg_flags=NLM_F_REQUEST, nlmsg_seq=2, nlmsg_pid=0}, {nfgen_family=AF_UNSPEC, version=NFNETLINK_V0, res_id=htons(2560)}]], iov_len=312}], msg_iovlen=1, msg_controllen=0, msg_flags=0}, 0) = 312
...
```

From the above `sendmsg` call observe the interpreted part: 
```
# Denotes the limit expression, message length, etc.
limit\x00\x00\x004\x00\x02\x80
# Following parts follow format: 
#   Length | Limit Field ID | Limit Field Data
# 12 | RATE  | 10
\x0c\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\n
# 12 | UNIT  | 1
\x0c\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x01
#  8 | BURST | 5   <-- here it is
\x08\x00\x03\x00\x00\x00\x00\x05
#  8 | TYPE  | 0 (NFT_LIMIT_PKTS)
\x08\x00\x04\x00\x00\x00\x00\x00
#  8 | FLAGS | 16
\x08\x00\x05\x00\x00\x00\x00x10
```

As far as I see, all of the data is passed to the kernel regardless of whether they are specified or not, so the ifs are not needed - if burst is not specified nftables will set burst to 5. I observe the same behaviour with the following struct:
```
# sed -n '52,69p' main.go 
	&expr.Dynset{
		SetID: set.ID,
		SrcRegKey: 1,
		SetName: set.Name,
		Operation: uint32(unix.NFT_DYNSET_OP_ADD),
		Exprs: []expr.Any{
			&expr.Connlimit{
				Count: 20,
				Flags: 1,
			},
			&expr.Limit{
				Type: expr.LimitTypePkts,
				Rate: 10,
				Unit: expr.LimitTimeSecond,
				Over: true,
			},
		},
	},
# go run main.go
filter forward
EXPRS: &{Count:20 Flags:1}
EXPRS: &{Type:0 Rate:10 Over:true Unit:1 Burst:5}
```

expr/expr.go
turekt: The `ExprsFromMsg` was previously [`exprsFromMsg` in `rule.go`](https://github.com/google/nftables/blob/3e042f75d78f94b3182304566f908f5bf7d7a7ad/rule.go#L218), but since it does not make sense to implement the function twice (it is needed to parse dynset expressions the same way as rule expressions) it was moved to the `expr` package and exported to be able to use it in both `rule.go` and `dynset.go`.

expr/limit.go
stapelberg: Thanks for the detailed reply. It took me a while to find this, but `nft` indeed defaults burst to 5, hidden in the parser: https://git.netfilter.org/nftables/tree/src/parser_bison.y?id=994bf5004b365904029f0fe8c2de587178583712#n3304

The Linux kernel itself also defaults to 5, which is likely why you’re seeing the `Burst:5` field when using the Go nftables package: https://github.com/torvalds/linux/blob/39c3c396f8131f3db454c80e0fcfcdc54ed9ec01/net/netfilter/nft_limit.c#L56

expr/expr.go
stapelberg: I’m concerned about exporting these and thereby making them part of the public API. Can you move the functionality into a package within the `internal/` directory instead please?

turekt: @stapelberg Understandable. I would be very happy to do it but unfortunately I am not sure how to move it without redesigning current source code placement.

The problem with moving the parsing functionality outside of the `expr` package is in the cyclic dependency that it creates between `expr` and `internal`.

Since the parsing logic extensively uses functions and structs from the `expr` package (`Unmarshal`, `Marshal`, `&expr.Notrack`, ...) moving it out of `expr` package would create the need to reference this (now moved) implementation from the same `expr` package while the implementation references `expr` package structs and functions. 

Let me try to depict this "graphically":
```
expr/dynset.go
  ^       |
  |       |
._|_________________________________________.
|uses expr structs and functions for parsing|
._|_________________________________________.
  |
  | CYCLE |
  |    .__|_____________________________________________________.
  |    |needs parsing_implementation to parse dynset expressions|
  |    .__|_____________________________________________________.
  |       |
  |       |
  |       v
internal/nfexpr/parsing_implementation.go
    ^
    |
  ._|____________________________________________________.
  |needs parsing_implementation to parse rule expressions|
  ._|____________________________________________________.
    |
    |
rule.go
```

I could theoretically overcome this by putting a part of the parsing implementation (the one which is not common for both `rule.go` and `dynset.go`?) in the newly created `internal` package and then force the `dynset.go` implementation to reference the parsing implementation in the `expr` package and `rule.go` to reference the parsing implementation in the newly created `internal` package where afterwards both reference the same (common) parsing implementation in the `expr` package, but this would still require me to export this parsing implementation in `expr` in order for it to be reachable from the newly created `internal` package.

Something like this:
```
expr/dynset.go (part1 parsing in Unmarshal)
         |
      .__|_____________________.
      |parse dynset expressions|
      .__|_____________________.
         |
         |
         v
expr/part2_of_parsing_implementation.go
       ^
       |
.______|________________________________________.
|parse rule expressions (cannot call if private)|
.______|________________________________________.
       |
       |
internal/nfexpr/part1_of_parsing_implementation.go
    ^
    |
  ._|____________________.
  |parse rule expressions|
  ._|____________________.
    |
    |
rule.go
```

Thinking twice, this approach does not make sense since the first part of rule parsing can just be written as a private function which would again need to call into `expr`, forcing the implementation in `expr` to be exported. The only advantage in this approach might be in exposing only a smaller part of the parsing implementation that you would not want to export.

Another approach I was thinking about was creating an `internal` package inside `expr` package but this is not possible since then the parsing implementation would be completely unavailable to `rule.go`.

Maybe I am just overlooking something. Let me think about other ways to approach this, but please feel free to ping here if you have any ideas.

stapelberg: Yeah, it’s a tricky situation. I think the following strategy could work, though:

1. Introduce a `internal/parseexprfunc` package which contains `var ParseFunc func(fam byte, ad *netlink.AttributeDecoder, b []byte) ([]interface{}, error)` (we need to use the empty interface instead of `expr.Any` to avoid a circular dependency).
2. In package `expr`, add:
    ```
   func init() { 
    parseexprfunc.ParseFunc = func(fam byte, ad *netlink.AttributeDecoder, b []byte) ([]interface{}, error) {
      exprs, err := exprsFromBytes(fam, ad, b)
      if err != nil {
        return nil, err
      }
      result := make([]interface{}, len(exprs))
      for idx, expr := range exprs {
        result[idx] = expr
      }
      return result, nil
    }
    ```
3. From `rule.go`, you can now call `parseexprfunc.ParseFunc()` and convert the opposite direction (from `interface{}` to `expr.Any`), and within package `expr`, you can call the non-exported `exprsFromBytes` for the dynset.

turekt: @stapelberg Thanks for the suggestion! I am not particularly happy with performing conversions back and forth but since I have no better approach to suggest, I have used your suggestion and implemented the `parseexprfunc` package.

stapelberg (approved):

stapelberg:
Thanks!

PR #174, opened 2022-08-08
Fix Meta unmarshal when is Source Register

methril:
The Meta nftables expression was not filling the Register and SourceRegister fields when unmarshalling.

Add a check for NFTA_META_SREG message when unmarshalling to fill the Meta fields.
Add Unit Test for source and destination unmarshall.


methril:
@stapelberg Could you please take a look to this PR?

stapelberg (approved):

Issue #176, opened 2022-08-27
Unable to create rule with verdict map


capnspacehook:
I've had success creating sets, verdict maps, and simple rules with this library, but I am unable to create one rule in particular:

```
table ip filter {
	map whalewall-ipv4-input-allow {
		type ipv4_addr . inet_proto . inet_service . ct_state : verdict
	}

	chain whalewall {
		ip daddr . ip protocol . tcp sport . ct state vmap @whalewall-ipv4-input-allow
	}
}
```

I used the netlink bytecode dumped by `nft` as a reference, but I am unable to create this rule for some reason. A minimal example of the Go code is here: https://go.dev/play/p/YmXKFtIeMwo.

I'm using Pop-OS with kernel version `5.19.0-76051900-generic`, Go 1.19, and nftables version `1.0.2-1ubuntu3`.

turekt:
Hi @capnspacehook,

your code works for me if I set the lookup expr `IsDestRegSet` to `true`:
```
&expr.Lookup{
  SourceRegister: 1,
  SetName:        ipv4InputMapName,
  DestRegister:   0,
  IsDestRegSet:   true,
},
```

capnspacehook:
Hey thanks for the tip @turekt that worked! In retrospect I probably should have thought of trying that.The reason I initially didn't however is because I created the rule using `nft` and used this library to load the rule and use that as my basis for creating the rule programmatically. However when loading this rule, `IsDestRegSet` of the final expression is set to `false`. Not a huge deal, but it appears unmarshalling logic has a small bug.

PR #178, opened 2022-08-29
IsDestRegSet unmarshaling fix

turekt:
Hi,

as noted in issue #176, the Lookup expression does not set the `IsDestRegSet` when unmarshaling the expr bytes. This is now fixed and a small test case is added.

Let me know what think.


stapelberg (approved):

stapelberg:
Thanks!

Issue #175, opened 2022-08-16
GetSetElements fails unexpected header type: got error, want unknown(2572)


Xniveres:
hey, i tried to get the size of a table, however GetSetElements returns an error:

code I've used:

```go
tables, _ := clientNFT.ListTables()
	for _, i := range tables {
		if i.Name == "main" {
			t, err := clientNFT.GetSetByName(i, "blackhole")
			if err != nil {
				fmt.Println(err.Error())
				return 0
			}
			fmt.Println("found table")
			fmt.Println(t.Name)
			data, err := clientNFT.GetSetElements(t)
			if err != nil {
				fmt.Println(err)
				return 0
			}
			return len(data)
		}
	}
```

output:

```go
found table
blackhole
unexpected header type: got error, want unknown(2572)
```

testing on go 1.18.3 on AlmaLinux 8.6

stapelberg:
Unfortunately I don’t have any experience with sets (cc @sbezverk who contributed set support).

If you can make your configuration work with the `nft` tool, can you compare what `nft` does and what the `nftables` package does? For an example, see https://github.com/google/nftables/blob/2eca00135732ff1a3b172d5ad68448294688290e/nftables_test.go#L270

That should hopefully make the problem relatively obvious.

turekt:
Hi,

I have managed to investigate this issue a bit better and I succeeded in reproducing this error when using nftables lasting connection. Here is example code:
```
func test() {
    c, _ := nftables.New(nftables.AsLasting())
    filter := &nftables.Table{
        Family: nftables.TableFamilyIPv4,
        Name:   "filter",
    }   
    c.AddTable(filter)
    c.Flush()
    testSet := &nftables.Set{
        Table:   filter,
        Name:    "test_set",
        KeyType: nftables.TypeIFName,
    }   
    c.AddSet(testSet, nil)
    c.Flush()

    c.GetSetByName(filter, "test_set")
    if _, err := c.GetSetElements(testSet); err != nil {
        panic(err)
    }   

    elements := []nftables.SetElement{{Key: []byte("wg1\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")}}
    if err := c.SetAddElements(testSet, elements); err != nil {
        panic(err)
    }   
    c.Flush()
}
```

[...]

Issue #103, opened 2020-07-13
Receive: netlink receive: recvmsg: no buffer space available


danopia:
Hello, I've been using this library to repeatedly 'replace' a table with ~50 rules, and one thing I've noticed is that certain patterns seem to trigger this cryptic message:

`Receive: netlink receive: recvmsg: no buffer space available`

The cryptic part is that it doesn't seem related to the overall size of the table, but rather the presence of certain rules (jumps, etc).

Even so, the flush seems to take effect anyway; the changes are reflected in the kernel after receiving the message.

I found some info on ENOBUFS here: https://netfilter.vger.kernel.narkive.com/cwzlgk8d/why-no-buffer-space-available and noticed that the golang netlink library has an option exposed to help the error: https://github.com/mdlayher/netlink/blob/c5f8ab79aa345dcfcf7f14d746659ca1b80a0ecc/conn.go#L437

I patched my local checkout of google/nftables/conn.go to include this line:

```go
        conn.SetOption(netlink.NoENOBUFS, true)
```
after:
```go
        defer conn.Close()
```

and it seems to fix the error.. I don't know enough at this point to understand the implications of using that option, or why it might not be desired, so I decided to start with a ticket instead of a PR.

Has anyone else seen this message?

stapelberg:
@mdlayher any thoughts?

mdlayher:
Unfortunately I don't have any more info. I've never run into this error in my work and only chose to implement the option because it appeared in a list in the man pages.

danopia:
Ok thanks, I'll see if I can capture and minify a repro for this. I assume without that there's basically nothing to look at 😄 

I am running my code on two different Debian releases, 10 and sid, and [see the error frequently on both](https://i.imgur.com/JXprmLY.png). (I've modified my calling code to consider `ENOBUFS` as a successful Flush(), instead of patching `NoENOBUFS` on, which might be a more reasonable workaround. Really not sure though.)

mdlayher:
I took a look at the mailing list thread you mentioned above and it seems to indicate that userspace cannot keep up with the kernel's messages.

When I wrote my netlink code, simplicity and correctness were the primary goals. I know there are some outstanding work items related to OS thread locking and etc that can slow things down, but since I haven't run into them in my use cases, it's low priority work for me.

You could try `DisableNSLockThread` in https://godoc.org/github.com/mdlayher/netlink#Config but I have no other advice at the moment unfortunately.

danopia:
[...]

PR #191, opened 2022-09-21
Netlink conn retry logic

turekt:
Hi,

I have introduced an additional function to the `conn.go` file which implements receive of netlink messages with retries. This should fix issue #175.

For some reason, in a lasting connection netlink sends a "no error" message with `NLM_F_CAPPED` flag set, and afterwards sends the actual message. Therefore, since there is nothing special to handle upon receiving the `NLM_F_CAPPED` + "no error" message, we can just drop the message and receive the next non-error message which should be the expected response.

Let me know what you think about this approach.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -15,14 +15,21 @@
 package nftables
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
+	"github.com/google/nftables/binaryutil"
 	"github.com/google/nftables/expr"
 	"github.com/mdlayher/netlink"
 	"github.com/mdlayher/netlink/nltest"
 	"golang.org/x/sys/unix"
 )
+
+// errMsgNoError is returned in cases when netlink returns an error
+// message with a status 0, which means that there is no error
+// https://github.com/mdlayher/netlink/blob/9a593f9dc1a92ae1c7c45a9fc663b7dd1e111eac/message.go#L285-L289
+var errMsgNoError = errors.New("no error")
 
 // A Conn represents a netlink connection of the nftables family.
 //
@@ -128,6 +135,48 @@ func (cc *Conn) netlinkConnUnderLock() (*netlink.Conn, netlinkCloser, error) {
 		return nil, nil, err
 	}
 	return nlconn, func() error { return nlconn.Close() }, nil
+}
+
+func receiveWithRetry(nlconn *netlink.Conn) ([]netlink.Message, error) {
+	if nlconn == nil {
+		return nil, fmt.Errorf("netlink conn is not initialized")
+	}
+
+	receive := func() ([]netlink.Message, error) {
+		var err error
+		reply, err := nlconn.Receive()
+		if err != nil {
+			return nil, err
+		}
+		if len(reply) == 0 {
+			return reply, nil
+		}
+		msg := reply[0]
+		if msg.Header.Type == netlink.Error {
+			if binaryutil.BigEndian.Uint32(msg.Data[:4]) != 0 {
+				return nil, fmt.Errorf("error delivered in message: %v", msg.Data)
+			}
+			return nil, errMsgNoError
+		}
+		return reply, nil
+	}
+
+	msgs, err := receive()
+	if err != nil {
+		// when "overflowing" counter objects netlink will return multiple
+		// "no error" messages (see TestCappedErrMsgOnObj in nftables_test.go)
+		// the loop is here to filter them all out
+		for err == errMsgNoError {
+			msgs, err := receive()
+			if err == nil {
+				return msgs, nil
+			}
+			if err != errMsgNoError {
+				return nil, err
+			}
+		}
+	}
+	return msgs, err
 }
 
 // CloseLasting closes the lasting netlink connection that has been opened using
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -2457,6 +2457,151 @@ func TestCreateUseAnonymousSet(t *testing.T) {
 	}
 }
 
+func TestCappedErrMsgOnObj(t *testing.T) {
+	c, newNS := openSystemNFTConn(t)
+	c, err := nftables.New(nftables.WithNetNSFd(int(newNS)), nftables.AsLasting())
+	if err != nil {
+		t.Fatalf("nftables.New() failed: %v", err)
+	}
+	defer cleanupSystemNFTConn(t, newNS)
+	c.FlushRuleset()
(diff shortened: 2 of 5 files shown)


conn.go
```diff
@@ -15,15 +15,22 @@
...
+	"github.com/google/nftables/binaryutil"
 	"github.com/google/nftables/expr"
 	"github.com/mdlayher/netlink"
 	"github.com/mdlayher/netlink/nltest"
 	"golang.org/x/sys/unix"
 )
 
+// errMsgNoError is returned in cases when netlink returns an error
+// message with a status 0, which means that there is no error
+// https://github.com/mdlayher/netlink/blob/main/message.go#L285-L289
```
stapelberg, line 31: Use the `y` keyboard shortcut on GitHub to turn this into a more stable URL: https://github.com/mdlayher/netlink/blob/9a593f9dc1a92ae1c7c45a9fc663b7dd1e111eac/message.go#L285-L289 :)

conn.go
```diff
@@ -130,6 +137,47 @@ func (cc *Conn) netlinkConnUnderLock() (*netlink.Conn, netlinkCloser, error) {
...
+
+	receive := func() ([]netlink.Message, error) {
+		var err error
+		reply, err := nlconn.Receive()
+		if err != nil {
+			return nil, err
+		}
+		if len(reply) > 0 {
+			msg := reply[0]
+			if got, want := msg.Header.Type, netlink.Error; got == want {
```
stapelberg, line 153: This can just be `if msg.Header.Type == netlink.Error {`

conn.go
```diff
@@ -130,6 +137,47 @@ func (cc *Conn) netlinkConnUnderLock() (*netlink.Conn, netlinkCloser, error) {
...
+		return nil, fmt.Errorf("netlink conn is not initialized")
+	}
+
+	receive := func() ([]netlink.Message, error) {
+		var err error
+		reply, err := nlconn.Receive()
+		if err != nil {
+			return nil, err
+		}
+		if len(reply) > 0 {
```
stapelberg, line 151: Indent the error case, not the normal case:

```go
if len(reply) == 0 {
  return reply, nil
}
msg := reply[0]
// …
```

nftables_test.go
```diff
@@ -2457,6 +2457,136 @@ func TestCreateUseAnonymousSet(t *testing.T) {
...
+	}
+
+	if err := c.Flush(); err != nil {
+		if got, want := fmt.Sprint(err), "no buffer space available"; !strings.Contains(got, want) {
+			t.Errorf("expected error %s, got %v", want, err)
+		}
+	}
+
+	// due to large number of counters
+	// netlink will return the "no error" message
```
stapelberg, line 2497: Do you happen to know the corresponding kernel code? Might be good to add a link for folks who want to understand this behavior in detail :)

stapelberg (commented):
Thanks for looking into the error handling! 

There are a couple of small things to change, but in general this seems like a good improvement 👍

nftables_test.go
turekt: I have rephrased and added additional information regarding this test case. It is not about the kernel code, it is related to the netlink go library which tries to peek for data that is too large to fit the socket read buffer. This results in not receiving the "no error" ack messages when expected and some of those messages end up as a response to the `GetObj` call done afterwards.

nftables_test.go
```diff
@@ -2457,6 +2457,151 @@ func TestCreateUseAnonymousSet(t *testing.T) {
...
+		// is too small to receive all ack messages sent by the kernel
+		//
+		// the Flush method tries to receive all acks after sending the messages
+		// https://github.com/google/nftables/blob/cbeb0fb1eccf9ef582c20982c72e73107d1898a5/conn.go#L185-L193
+		//
+		// buffer peeking implemented by the netlink go library fails with a "no buffer space available"
+		// https://github.com/mdlayher/netlink/blob/7fa043dcb6f27ed7e084dc90ddf5b6d4092478a9/conn_linux.go#L123-L127
+		// this results in part of "non-received" acknowledgments to end up being received
+		// in the following c.GetObj(counters[0]) call
+		//
```
stapelberg, line 2501: Thanks for the additional description and details.

This makes me wonder: could we instead fix this in github.com/mdlayher/netlink by handling the “no buffer space” error and enlarging the buffer, as that code already tries to do? Would that make this entire PR unnecessary?

turekt: Good thinking but unfortunately no. Even if we would propose a PR fix that would enlarge the buffer, we would still have the issue of the "no error" messages being sent in GetSetElements as initially reported in issue #175 (and including other functions using conn.Receive without retries). Note that the "no error" messages seen in the issue are not caused by the "no buffer space" error. The "no error" messages (basically acks) are a result of using the nftables lasting connection feature. At which point and why we receive these acks in the lasting connection is [still unknown to me](https://github.com/google/nftables/issues/175#issuecomment-1233191956).

What I am doing with this test is only mimicking this behaviour with GetObj function since [the comment reporting the issue with GetObj](https://github.com/google/nftables/issues/175#issuecomment-1252691215) was not specific on how to reproduce the issue so I was not able to write that specific test other than reproducing the "no buffer space" error and forcing acknowledgements as a response for the GetObj function. I know that it seems a bit hacky, but it still shows as a good test for the retry logic.

Let me also note that "no buffer space" error was already discussed once in issue #103 but the issue was closed as stale.

stapelberg: > Good thinking but unfortunately no. Even if we would propose a PR fix that would enlarge the buffer, we would still have the issue of the "no error" messages being sent in GetSetElements as initially reported in issue #175 (and including other functions using conn.Receive without retries). Note that the "no error" messages seen in the issue are not caused by the "no buffer space" error. The "no error" messages (basically acks) are a result of using the nftables lasting connection feature. 

I think these two issues might best be fixed separately — the “no buffer space” one in mdlayher/netlink, and the acknowledgement issue in google/nftables.

> At which point and why we receive these acks in the lasting connection is [still unknown to me](https://github.com/google/nftables/issues/175#issuecomment-1233191956).

I think it would be best if we actually understood what the issue with the acks is. It still sounds to me like we’re requesting more acks than we should, or something along these lines.

While your PR probably works, it seems to me like it’s papering over the issue.

turekt: Agreed, thanks for the input. I have introduced a different approach in PR #194. I will try to address the "no buffer" error separately in netlink library.

PR #194, opened 2022-09-29
Netlink conn retry logic v2

turekt:
Hi,

I am opening a version 2 PR for review and discussion, related to PR #191. I have further investigated the error from issue #175 and I understand why the issue is happening for the nftables lasting connection. The root cause for this issue is that it seems that the acknowledgments for messages are being sent under specific conditions.

By observing the following debug output:
```
// Batch message init - add table filter, request ack
nl: send msgs: {Header:{Length:20 Type:unknown(16) Flags:request Sequence:1655472395 PID:5277} Data:[0 0 0 10]}
nl: send msgs: {Header:{Length:40 Type:unknown(2560) Flags:request|acknowledge|0x400 Sequence:1655472396 PID:5277} Data:[2 0 0 0 11 0 1 0 102 105 108 116 101 114 0 0 8 0 2 0 0 0 0 0]}
nl: send msgs: {Header:{Length:20 Type:unknown(17) Flags:request Sequence:1655472397 PID:5277} Data:[0 0 0 10]}
// Acknowledge sequence [12 129 172 98] = 1655472396, add table message
nl: recv: {Header:{Length:36 Type:error Flags:0x100 Sequence:1655472396 PID:5277} Data:[0 0 0 0 40 0 0 0 0 10 5 4 12 129 172 98 157 20 0 0]}

// Batch message init - add set test_set, request ack
nl: send msgs: {Header:{Length:20 Type:unknown(16) Flags:request Sequence:1655472398 PID:5277} Data:[0 0 0 10]}
nl: send msgs: {Header:{Length:92 Type:unknown(2569) Flags:request|acknowledge|0x400 Sequence:1655472399 PID:5277} Data:[2 0 0 0 11 0 1 0 102 105 108 116 101 114 0 0 13 0 2 0 116 101 115 116 95 115 101 116 0 0 0 0 8 0 3 0 0 0 0 0 8 0 4 0 0 0 0 41 8 0 5 0 0 0 0 16 8 0 10 0 0 0 0 1 10 0 13 0 0 4 1 0 0 0 0 0]}
nl: send msgs: {Header:{Length:20 Type:unknown(17) Flags:request Sequence:1655472400 PID:5277} Data:[0 0 0 10]}
// Acknowledge sequence [15 129 162 98] = 1655472399, add set message
nl: recv: {Header:{Length:36 Type:error Flags:0x100 Sequence:1655472399 PID:5277} Data:[0 0 0 0 92 0 0 0 9 10 5 4 15 129 172 98 157 20 0 0]}

// Get set by name test_set, not batch
nl: send msgs: {Header:{Length:48 Type:unknown(2570) Flags:request|acknowledge Sequence:1655472401 PID:5277} Data:[2 0 0 0 11 0 1 0 102 105 108 116 101 114 0 0 13 0 2 0 116 101 115 116 95 115 101 116 0 0 0 0]}
// Send test_set data
nl: recv: {Header:{Length:92 Type:unknown(2569) Flags:0 Sequence:1655472401 PID:5277} Data:[2 0 0 4 11 0 1 0 102 105 108 116 101 114 0 0 13 0 2 0 116 101 115 116 95 115 101 116 0 0 0 0 12 0 16 0 0 0 0 0 0 0 0 1 8 0 4 0 0 0 0 41 8 0 5 0 0 0 0 16 10 0 13 0 0 4 1 0 0 0 0 0 4 0 9 0]}

// Get set elements test_set, not batch
nl: send msgs: {Header:{Length:48 Type:unknown(2573) Flags:request|acknowledge|0x300 Sequence:1655472402 PID:5277} Data:[2 0 0 0 11 0 1 0 102 105 108 116 101 114 0 0 13 0 2 0 116 101 115 116 95 115 101 116 0 0 0 0]}
// Acknowledge sequence [17 129 172 97] = 1655472401, get set by name message ! out-of-sync !
nl: recv: {Header:{Length:36 Type:error Flags:0x100 Sequence:1655472401 PID:5277} Data:[0 0 0 0 48 0 0 0 10 10 5 0 17 129 172 98 157 20 0 0]}
// Send test_set elements data, as multi, no ack returned
nl: recv: {Header:{Length:80 Type:unknown(2572) Flags:multi Sequence:1655472402 PID:5277} Data:[2 0 0 4 11 0 1 0 102 105 108 116 101 114 0 0 13 0 2 0 116 101 115 116 95 115 101 116 0 0 0 0 32 0 3 0 28 0 1 0 24 0 1 0 20 0 1 0 119 103 49 0 0 0 0 0 0 0 0 0 0 0 0 0]}
nl: recv: {Header:{Length:52 Type:unknown(2572) Flags:multi Sequence:1655472402 PID:5277} Data:[2 0 0 4 11 0 1 0 102 105 108 116 101 114 0 0 13 0 2 0 116 101 115 116 95 115 101 116 0 0 0 0 4 0 3 0]}
nl: recv: {Header:{Length:20 Type:done Flags:multi Sequence:1655472402 PID:5277} Data:[0 0 0 0]}

// Batch message init - add set element wg1
nl: send msgs: {Header:{Length:20 Type:unknown(16) Flags:request Sequence:1655472403 PID:5277} Data:[0 0 0 10]}
nl: send msgs: {Header:{Length:88 Type:unknown(2572) Flags:request|acknowledge|0x400 Sequence:1655472404 PID:5277} Data:[2 0 0 0 13 0 2 0 116 101 115 116 95 115 101 116 0 0 0 0 8 0 4 0 0 0 0 1 11 0 1 0 102 105 108 116 101 114 0 0 32 0 3 128 28 0 1 128 24 0 1 128 20 0 1 0 119 103 49 0 0 0 0 0 0 0 0 0 0 0 0 0]}
nl: send msgs: {Header:{Length:20 Type:unknown(17) Flags:request Sequence:1655472405 PID:5277} Data:[0 0 0 10]}
// Acknowledge sequence [20 129 172 98] = 1655472404, add set element message
nl: recv: {Header:{Length:36 Type:error Flags:0x100 Sequence:1655472404 PID:5277} Data:[0 0 0 0 88 0 0 0 12 10 5 4 20 129 172 98 157 20 0 0]}
```

Similar behaviour is observed with `GetObj` - no acks are given in case `NLM_F_MULTI` is set in response. I have confirmed this behaviour by checking the kernel code: https://github.com/torvalds/linux/blob/7e062cda7d90543ac8c7700fc7c5527d0c0f22ad/net/netlink/af_netlink.c#L2387-L2390, if `NLM_F_DUMP` is set, kernel runs the dump function which will always skip sending acks.

This PR introduces a different fix by checking if the connection is lasting and verifying if the send message is flagged as `NLM_F_DUMP` to determine if we can expect an ack message or not.

Let me know what you think about this approach.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -15,9 +15,11 @@
 package nftables
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
+	"github.com/google/nftables/binaryutil"
 	"github.com/google/nftables/expr"
 	"github.com/mdlayher/netlink"
 	"github.com/mdlayher/netlink/nltest"
@@ -128,6 +130,52 @@ func (cc *Conn) netlinkConnUnderLock() (*netlink.Conn, netlinkCloser, error) {
 		return nil, nil, err
 	}
 	return nlconn, func() error { return nlconn.Close() }, nil
+}
+
+func receiveAckAware(nlconn *netlink.Conn, sentMsgFlags netlink.HeaderFlags) ([]netlink.Message, error) {
+	if nlconn == nil {
+		return nil, errors.New("netlink conn is not initialized")
+	}
+
+	// first receive will be the message that we expect
+	reply, err := nlconn.Receive()
+	if err != nil {
+		return nil, err
+	}
+
+	if (sentMsgFlags & netlink.Acknowledge) == 0 {
+		// we did not request an ack
+		return reply, nil
+	}
+
+	if (sentMsgFlags & netlink.Dump) == netlink.Dump {
+		// sent message has Dump flag set, there will be no acks
+		// https://github.com/torvalds/linux/blob/7e062cda7d90543ac8c7700fc7c5527d0c0f22ad/net/netlink/af_netlink.c#L2387-L2390
+		return reply, nil
+	}
+
+	// Dump flag is not set, we expect an ack
+	ack, err := nlconn.Receive()
+	if err != nil {
+		return nil, err
+	}
+
+	if len(ack) == 0 {
+		return nil, errors.New("received an empty ack")
+	}
+
+	msg := ack[0]
+	if msg.Header.Type != netlink.Error {
+		// acks should be delivered as NLMSG_ERROR
+		return nil, fmt.Errorf("expected header %v, but got %v", netlink.Error, msg.Header.Type)
+	}
+
+	if binaryutil.BigEndian.Uint32(msg.Data[:4]) != 0 {
+		// if errno field is not set to 0 (success), this is an error
+		return nil, fmt.Errorf("error delivered in message: %v", msg.Data)
+	}
+
+	return reply, nil
 }
 
 // CloseLasting closes the lasting netlink connection that has been opened using
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -1758,6 +1758,7 @@ func TestGetObjReset(t *testing.T) {
 		nil,
 		[]netlink.Message{netlink.Message{Header: netlink.Header{Length: 0x64, Type: 0xa12, Flags: 0x802, Sequence: 0x9acb0443, PID: 0xde9}, Data: []uint8{0x2, 0x0, 0x0, 0x10, 0xb, 0x0, 0x1, 0x0, 0x66, 0x69, 0x6c, 0x74, 0x65, 0x72, 0x0, 0x0, 0xa, 0x0, 0x2, 0x0, 0x66, 0x77, 0x64, 0x65, 0x64, 0x0, 0x0, 0x0, 0x8, 0x0, 0x3, 0x0, 0x0, 0x0, 0x0, 0x1, 0x8, 0x0, 0x5, 0x0, 0x0, 0x0, 0x0, 0x1, 0x1c, 0x0, 0x4, 0x0, 0xc, 0x0, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x4, 0x61, 0xc, 0x0, 0x2, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x9, 0xc, 0x0, 0x6, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x2}}},
 		[]netlink.Message{netlink.Message{Header: netlink.Header{Length: 0x14, Type: 0x3, Flags: 0x2, Sequence: 0x9acb0443, PID: 0xde9}, Data: []uint8{0x0, 0x0, 0x0, 0x0}}},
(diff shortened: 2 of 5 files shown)


conn.go
```diff
@@ -130,6 +132,58 @@ func (cc *Conn) netlinkConnUnderLock() (*netlink.Conn, netlinkCloser, error) {
...
+	}
+
+	// first receive will be the message that we expect
+	reply, err := nlconn.Receive()
+	if err != nil {
+		return nil, err
+	}
+
+	if !cc.lasting {
+		// in non lasting connection there are no acks being sent,
```
stapelberg, line 147: I agree that a non-lasting connection cannot be confused by out of sync acks, but shouldn’t we still read the acks so that we do error checking (line 174 below), even if we open a new connection afterwards?

conn.go
```diff
@@ -130,6 +132,58 @@ func (cc *Conn) netlinkConnUnderLock() (*netlink.Conn, netlinkCloser, error) {
...
+		return reply, nil
+	}
+
+	// Dump flag is not set, we expect an ack
+	ack, err := nlconn.Receive()
+	if err != nil {
+		return reply, err
+	}
+
+	if len(ack) == 0 {
```
stapelberg, line 169: Isn’t len(ack) == 0 an error case? I.e. we requested an ack but didn’t get one?

stapelberg (commented):
Thanks for revising your approach, this looks much better! :)

conn.go
turekt: Agreed.

conn.go
turekt: Thanks, I was not sure what is the best approach to handle this since we've already got the reply that we need at this point and acks are not something that we handle other than receive and "ignore". In any case, we can notify that ack was empty but still return the original reply and let the user decide what to do with it.

conn.go
```diff
@@ -130,6 +132,52 @@ func (cc *Conn) netlinkConnUnderLock() (*netlink.Conn, netlinkCloser, error) {
...
+		return nil, err
+	}
+
+	if (sentMsgFlags & netlink.Acknowledge) == 0 {
+		// we did not request an ack
+		return reply, nil
+	}
+
+	if (sentMsgFlags & netlink.Dump) == netlink.Dump {
+		// sent message has Dump flag set, there wil be no acks
```
stapelberg, line 152: typo: wil → will

conn.go
stapelberg: If the user is supposed to decide, we need to make the returned error a sentinel so folks can use errors.Is/errors.As. 

But, I don’t think it’s a good choice here to let the user decide. An empty ack sounds like a bug in our library and/or in the kernel, not something that one group of users might want while another group might not want.

I would favor returning just an error, and if we get reports of problems with that approach we can revisit.

turekt: Done.

stapelberg (approved):

stapelberg:
Thanks very much!

PR #195, opened 2022-10-12
add int32 and string types to alignedbuff

joewilliams:
We use `xt.Unknown` to use xtables matches unsupported directly by this library. To create the correct struct layout we use alignedbuff and binaryutil but a couple of types we need are missing. This PR adds those types and tests. This PR also works in our custom marshaling code for these xtable structs.

### DIFF
--- a/binaryutil/binaryutil.go
+++ b/binaryutil/binaryutil.go
@@ -16,6 +16,7 @@
 package binaryutil
 
 import (
+	"bytes"
 	"encoding/binary"
 	"unsafe"
 )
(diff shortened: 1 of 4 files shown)


binaryutil/binaryutil.go
```diff
@@ -102,3 +103,22 @@ func (bigEndian) Uint32(b []byte) uint32 {
 func (bigEndian) Uint64(b []byte) uint64 {
 	return binary.BigEndian.Uint64(b)
 }
+
+// For dealing with types not supported by the encoding/binary interface
```
stapelberg, line 107: add a blank line after this comment to avoid it becoming the documentation of PutInt32.

binaryutil/binaryutil_test.go
```diff
@@ -107,3 +107,43 @@ func TestBigEndian(t *testing.T) {
...
+			expected:  []byte{0x74, 0x65, 0x73, 0x74},
+			expectedv: "test",
+			actual:    PutString("test"),
+			unmarshal: func(b []byte) interface{} { return String(b) },
+		},
+		{
+			name:      "NullTerminatedString",
+			expected:  []byte{0x74, 0x65, 0x73, 0x74, 0x00},
+			expectedv: "test",
+			actual:    PutNullTerminatedString("test"),
```
stapelberg, line 137: This function does not seem to exist. Can you fix the failing test please?

stapelberg (changes requested):

joewilliams:
@stapelberg should be fixed up, thank you!

stapelberg:
Thanks!

PR #199, opened 2022-10-24
Support set element counters

joewilliams:
This PR adds the ability to have counters on each set element.

This is equivalent to something like:
```
nft add set ip filter test {type ipv4_addr \; counter \;}
```

Afterward each element added to the set will have a counter associated with it:
```
	set ipv4_blocklist {
		type ipv4_addr
		flags interval
		counter
		elements = { 192.0.2.0/24 counter packets 60 bytes 3743, 198.51.100.1 counter packets 5 bytes 220,
			     203.0.113.100 counter packets 11 bytes 464 }
	}
```

### DIFF
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -2642,6 +2642,49 @@ func TestIP6SetAddElements(t *testing.T) {
 	}
 }
 
+func TestCreateUseCounterSet(t *testing.T) {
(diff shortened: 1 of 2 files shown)


stapelberg (approved):

stapelberg:
Thanks!

Issue #200, opened 2022-10-28
how to set flowtable offload


gzm314:
@stapelberg libnftnl supports setting flowtable offload, how can I implement offload through this project?

stapelberg:
Sorry, I don’t know. Maybe someone else knows. Otherwise you’ll have to figure it out. Maybe `nft --debug=all` prints enough info to put you on the right path.

gzm314:
nice work, thanks :)

PR #201, opened 2022-10-30
Flowtables implementation

turekt:
Hi,

I have added the flowtable and flow offload expression implementation to the library as requested in issue #200. Test cases demonstrate how one can implement flowtable offload using this implementation. I think this covers flowtable features completely but if something is missing I can add additional features.

Let me know what you think.


stapelberg (approved):

stapelberg:
Nice, thanks!

Issue #202, opened 2022-11-01
How to add the anonymous set of hours ? 


ymm135:
@turekt Hi. I'm having a problem with the go nftables api. i want to add the set of hours, i can use nft to set successfully.

The rule that I want to generate is :
`rules.nft`  
```shell
table ip filter {

        chain FORWARD {
                type filter hook forward priority filter; policy accept;
                meta hour { "16:00:00"-"18:00:00","20:00:00"-"22:00:00" } accept
        }
}
```

When I run your ruleset through `nft --debug=netlink -f rules.nft`, I get:
```shell
ip (null) (null) use 0 type filter hook forward prio 0 packets 0 bytes 0
__set%d filter 7 size 5
__set%d filter 0
	element 00000000  : 1 [end]	element 80700000  : 0 [end]	element a18c0000  : 1 [end]	element c0a80000  : 0 [end]	element e1c40000  : 1 [end]
ip filter FORWARD
  [ meta load hour => reg 1 ]
  [ byteorder reg 1 = hton(reg 1, 4, 4) ]
  [ lookup reg 1 set __set%d ]
  [ immediate reg 0 accept ]
```

my code is :
```go
package main

import (
	"bytes"
	"encoding/binary"
	"fmt"
	"runtime"
	"strings"
	"time"

	"github.com/google/nftables"
	"github.com/google/nftables/expr"
	"github.com/vishvananda/netns"
)

const (
	NFT_META_TIME_NS   = 30
	NFT_META_TIME_DAY  = 31
	NFT_META_TIME_HOUR = 32
)

// the running kernel in a separate network namespace.
// cleanupSystemNFTConn() must be called from a defer to cleanup
// created network namespace.
func openSystemNFTConn() (*nftables.Conn, netns.NsHandle) {
	runtime.LockOSThread()

	// init pid
	const init_pid = 1
	ns, err := netns.GetFromPid(init_pid)

	if err != nil {
		fmt.Sprintln("GetFromPid err", err.Error())
	}

	c, err := nftables.New(nftables.WithNetNSFd(int(ns)))

	if err != nil {
		fmt.Sprintln("nftables.New() failed", err.Error())
	}
	return c, ns
}

func cleanupSystemNFTConn(newNS netns.NsHandle) {
	defer runtime.UnlockOSThread()

	if err := newNS.Close(); err != nil {
		fmt.Printf("newNS.Close() failed: %v", err)
	}
}

// time
func metaTime(timeStr string) []byte {
[...]

PR #203, opened 2022-11-01
Fix size for TypeTimeHour

turekt:
Hi,

I have made a small correction in `set.go` for `TypeTimeHour`. The size of the type is set to 8 via `Bytes` field which is incorrect. The correct size should be 4. This is also confirmed in nftables code here: https://git.netfilter.org/nftables/tree/src/meta.c?id=c8a3c669499d169fef8c1e89b8d2d909e5ecd023#n600. The size change was introduced with the following commit: https://git.netfilter.org/nftables/commit/src/meta.c?id=4e1abfc552170d6db5c511634a29918e64c1b51b.

This PR should also fix issue #202.


stapelberg (approved):

stapelberg:
Thanks!

Issue #204, opened 2022-11-03
How to reject packets?


elmeyer:
Apologies if this is the wrong place to ask; I couldn't find any C code that does what I want to do either, and I'm not even sure this is possible via NFNETLINK.

I can execute the following series of `nft` commands to add a table, prerouting filter chain and rule rejecting any TCP packets to port 8080:

```
nft add table inet test

nft add chain inet test test-in { type filter hook prerouting priority 0 \; }

nft add rule inet test test-in ip saddr 0.0.0.0/0 ip daddr 0.0.0.0/0 tcp dport 8080 reject
```

`nft list table inet test` then shows the following:

```
table inet test {
	chain test-in {
		type filter hook prerouting priority filter; policy accept;
		ip saddr 0.0.0.0/0 ip daddr 0.0.0.0/0 tcp dport 8080 reject with icmp port-unreachable
	}
}
```

I have attached some sample code at the bottom. When called without any flag, the code lists all rules in all chains. When I run this code after executing the series of `nft` commands above, I get the following output:

```
([]*nftables.Rule) (len=1 cap=1) {
 (*nftables.Rule)(0x4000132300)({
  Table: (*nftables.Table)(0x400013e0e0)({
   Name: (string) (len=4) "test",
   Use: (uint32) 0,
   Flags: (uint32) 0,
   Family: (nftables.TableFamily) 1
  }),
  Chain: (*nftables.Chain)(0x4000170100)({
   Name: (string) (len=7) "test-in",
   Table: (*nftables.Table)(<nil>),
   Hooknum: (*nftables.ChainHook)(<nil>),
   Priority: (*nftables.ChainPriority)(<nil>),
   Type: (nftables.ChainType) "",
   Policy: (*nftables.ChainPolicy)(<nil>)
  }),
  Position: (uint64) 0,
  Handle: (uint64) 2,
  Flags: (uint32) 0,
  Exprs: ([]expr.Any) (len=12 cap=12) {
   (*expr.Meta)(0x4000138554)({
    Key: (expr.MetaKey) 15,
    SourceRegister: (bool) false,
    Register: (uint32) 1
   }),
   (*expr.Cmp)(0x400013e140)({
    Op: (expr.CmpOp) 0,
    Register: (uint32) 1,
    Data: ([]uint8) (len=1 cap=1) {
     00000000  02                                                |.|
    }
   }),
   (*expr.Payload)(0x400012e0f0)({
    OperationType: (expr.PayloadOperationType) 0,
[...]

PR #206, opened 2022-11-07
Reject expression fix

turekt:
Hi,

as pointed out in issue #204, the library is not unmarshaling reject expressions. This is due to a missing `case` in `exprsFromBytes`. This commit should also resolve the aformentioned issue.


stapelberg (approved):

stapelberg:
Thanks!

PR #207, opened 2022-11-21
Fix: add NFTA_RULE_COMPAT attribute

zhaofeng0019:
xt_matches or xt_targets like xt_tcpudp may have specific compat policy and if not set flush rule will error with EINVAL
according to https://elixir.bootlin.com/linux/v3.13/source/net/netfilter/x_tables.c#L563

Signed-off-by: xiaoff <zhaofeng_0019@163.com>

### DIFF
--- a/compat_policy.go
+++ b/compat_policy.go
@@ -1,0 +1,89 @@
+package nftables
+
+import (
+	"fmt"
+
+	"github.com/google/nftables/expr"
+	"golang.org/x/sys/unix"
+)
+
+const nft_RULE_COMPAT_F_INV uint32 = (1 << 1)
+const nft_RULE_COMPAT_F_MASK uint32 = nft_RULE_COMPAT_F_INV
+
+// Used by xt match or target like xt_tcpudp to set compat policy between xtables and nftables
+// https://elixir.bootlin.com/linux/v5.12/source/net/netfilter/nft_compat.c#L187
+type compatPolicy struct {
+	Proto uint32
+	Flag  uint32
+}
+
+var xtMatchCompatMap map[string]*compatPolicy = map[string]*compatPolicy{
+	"tcp": {
+		Proto: unix.IPPROTO_TCP,
+	},
+	"udp": {
+		Proto: unix.IPPROTO_UDP,
+	},
+	"udplite": {
+		Proto: unix.IPPROTO_UDPLITE,
+	},
+	"tcpmss": {
+		Proto: unix.IPPROTO_TCP,
+	},
+	"sctp": {
+		Proto: unix.IPPROTO_SCTP,
+	},
+	"osf": {
+		Proto: unix.IPPROTO_TCP,
+	},
+	"ipcomp": {
+		Proto: unix.IPPROTO_COMP,
+	},
+	"esp": {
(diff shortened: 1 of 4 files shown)


stapelberg:
cc @thediveo — does this look good to you?

thediveo:
LGTM -- the only thing I notice while I'm not an expert in this area: is this something only specified by an API user or can this information also be returned when retrieving corresponding data?

zhaofeng0019:
> LGTM -- the only thing I notice while I'm not an expert in this area: is this something only specified by an API user or can this information also be returned when retrieving corresponding data?

so maybe make it a private field and auto check exprs in the rule when flush it ?

zhaofeng0019:
@thediveo @stapelberg  new commit need to be review, make compat_policy transparent to user

zhaofeng0019:
so is there someone to see this?

thediveo:
Not so much from my side, but I think @stapelberg?

compat_policy.go
```diff
@@ -0,0 +1,89 @@
+package nftables
+
+import (
+	"fmt"
+
+	"github.com/google/nftables/expr"
+	"golang.org/x/sys/unix"
+)
+
+const NFT_RULE_COMPAT_F_INV uint32 = (1 << 1)
```
stapelberg, line 10: Can you make these two constants unexported please? (E.g. by lower-casing the “nft”)

We should avoid adding anything to the public API surface of the nftables package unless users really need it :)

compat_policy_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"testing"
+
+	"github.com/google/nftables/expr"
+	"github.com/google/nftables/xt"
+	"golang.org/x/sys/unix"
+)
+
+func TestGetCompatPolicy(t *testing.T) {
```
stapelberg, line 11: Can you add another test that installs an nftables rule which requires a compat policy?

compat_policy.go
```diff
@@ -0,0 +1,89 @@
...
+const NFT_RULE_COMPAT_F_MASK uint32 = NFT_RULE_COMPAT_F_INV
+
+// Used by xt match or target like xt_tcpudp to set compat policy between xtables and nftables
+// https://elixir.bootlin.com/linux/v3.13/source/net/netfilter/nft_compat.c#L126
```
stapelberg, line 14: Is there any reason you’re linking to v3.13 specifically? That’s quite an old Linux version by now. Could you link to the latest stable one instead?

stapelberg (changes requested):

compat_policy.go
zhaofeng0019: because 3.13 is the first version that kernel bring nftables into it. so think about compat, i usually use this version, but change kernel version is ok, because it actually don't change

compat_policy_test.go
zhaofeng0019: done

compat_policy.go
zhaofeng0019: done

compat_policy_test.go
zhaofeng0019: add now

stapelberg (approved):

Issue #209, opened 2022-12-10
Alignment issues on 32-bit archs: TestAlignedBuff32 & TestAlignedBuffInt32 failures


CyrilBrulebois:
Hi,

Seen on armel, armhf, and i386 via autopkgtest on the https://ci.debian.net/ infrastructure, i.e. all 32-bit architectures:

```
=== RUN   TestAlignedBuff32
    alignedbuff_test.go:107: alignment padding failed: 12
--- FAIL: TestAlignedBuff32 (0.00s)
```

and:

```
=== RUN   TestAlignedBuffInt32
    alignedbuff_test.go:218: alignment padding failed: 12
--- FAIL: TestAlignedBuffInt32 (0.00s)
```

(I've added the actual length by adjusting the source, 16 is expected in both places.)

I'm happy to try anything you'd like me to; this issue can be easily reproduced locally by building the `golang-github-google-nftables` package within an i386 development chroot (Debian unstable), e.g. on an amd64 system.

stapelberg:
cc @thediveo who contributed alignedbuff

thediveo:
I suspect my test to be (obviously) incorrectly done, I need to test the test next time. Somehow the 32bit archs return to exact revenge on me. Unfortunately, I need to setup a test VM or similar first.

CyrilBrulebois:
If you want to push a branch, commits, patches, etc. somewhere, I'm happy to run tests and report back if that lowers the barrier for you.

thediveo:
thank you very much! but I need to first explore this, as Go behaves differently from the C compilers in some aspects of 64bit versus 32bit architectures. nftables need to use the C alignment rules...

thediveo:
I've now set up a Debian 11 i386 VM on kvm/qemu ... the only bummer is that there's no 32bit versions of VSCode anymore. I can reproduce using go version 1.19.4 on linux/386.

CyrilBrulebois:
With your commit, it builds and tests fine at least on amd64 and i386 indeed, thanks! Would like me to run some tests on arm* systems (I have access to a few porter boxes via my Debian hat, including armel and armhf).

thediveo:
yes, please, if that is possible.

CyrilBrulebois:
Confirmed building and testing fine with 318b4afcdd40d17a3dd2f583e79586373be1626d on armel and armhf, in addition to amd64 and i386.

PR #211, opened 2022-12-10
alignedbuff: fix alignment test issue on 32-bit machines

thediveo:
- fixes issue #209 where two unit tests for alignedbuff were incorrectly calculating the expected marshalled data length on 32bit machines (whereas actual padding/alignment itself was done correctly).
- adds documentation reference to kernel's xtables.h UAPI regarding alignment.

### DIFF
--- a/alignedbuff/alignedbuff_test.go
+++ b/alignedbuff/alignedbuff_test.go
@@ -103,7 +103,15 @@ func TestAlignedBuff32(t *testing.T) {
 
 	b := NewWithData(b0.data)
 
-	if len(b0.Data()) != 4*4 {
+	// Sigh. The Linux kernel expects certain nftables payloads to be padded to
+	// the uint64 next alignment. Now, on 64bit platforms this will be a 64bit
+	// alignment, yet on 32bit platforms this will be a 32bit alignment. So, we
+	// should calculate the expected data length here separately from our
+	// implementation to be fail safe! However, this might be rather a recipe
+	// for a safe fail...
+	expectedlen := 2*(uint32AlignMask+1) + (uint64AlignMask + 1)
+
+	if len(b0.Data()) != expectedlen {
 		t.Fatalf("alignment padding failed")
 	}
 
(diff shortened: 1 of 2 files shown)


alignedbuff/alignedbuff_test.go
```diff
@@ -103,7 +103,15 @@ func TestAlignedBuff32(t *testing.T) {
 
 	b := NewWithData(b0.data)
 
-	if len(b0.Data()) != 4*4 {
+	// Sigh. The Linux kernel expects certain nftables payloads to be padded to
+	// the uint64 next alignment. Now, on 64bit platforms this will be a 64bit
+	// alignment, yet on 32bit platforms this will be a 32bit alignment. So, we
+	// should calculate the expected data length here separately from our
+	// implementation to be fail safe! However, this might be rather a receipe
```
stapelberg, line 110: here and below: receipe → recipe

alignedbuff/alignedbuff_test.go
```diff
@@ -103,7 +103,15 @@ func TestAlignedBuff32(t *testing.T) {
 
 	b := NewWithData(b0.data)
 
-	if len(b0.Data()) != 4*4 {
+	// Sigh. The Linux kernel expects certain nftables payloads to be padded to
+	// the uint64 next alignment. Now, on 64bit platforms this will be a 64bit
```
stapelberg, line 107: Doesn’t the Linux kernel expect payloads to be padded to **`int`** alignment (not uint64)? The size of an `int` differs between 32-bit and 64-bit as you outlined.

stapelberg (changes requested):
Can you update the commit message to be more descriptive please? Maybe something along the lines of “alignedbuff: fix alignment issue on 32-bit machines”

I can merge this once @CyrilBrulebois reports back whether it works on arm.

alignedbuff/alignedbuff_test.go
thediveo: fixed in both places; thank you for spotting!

alignedbuff/alignedbuff_test.go
thediveo: I (re) checked with x_tables.h (https://elixir.bootlin.com/linux/v5.17.7/source/include/uapi/linux/netfilter/x_tables.h#L93) and added that documentation reference to the package comment of `alignedbuff` so others can quickly find and check my reasoning. The alignment is done based on the alignment of a struct consisting of unsigned ints of 8, 16, 32, 64bits. I cross-checked with a tiny C program on amd64 as well as i386 and it gave me the same alignment results as the Go compiler on each architecture respectively: 64bits and 32bits alignments.

In consequence, my calculation of the expected marshalled data size was incorrect for 32bit platforms and this patch fixes my original incorrect expectation. The implemented alignment itself was and still is correct.

thediveo:
better?

stapelberg (approved):

stapelberg:
Thanks!

Issue #213, opened 2023-01-12
GetRules lost  expr.Masq


junhwong:
test rule: nft add rule inet nat mychain ip daddr != 224.0.0.0/4 counter masquerade

rule exprs
```

a-typ: *expr.Meta
a-val: &{Key:15 SourceRegister:false Register:1}
a-typ: *expr.Cmp
a-val: &{Op:0 Register:1 Data:[2]}
a-typ: *expr.Payload
a-val: &{OperationType:0 DestRegister:1 SourceRegister:0 Base:1 Offset:16 Len:4 CsumType:0 CsumOffset:0 CsumFlags:0}
a-typ: *expr.Bitwise
a-val: &{SourceRegister:1 DestRegister:1 Len:4 Mask:[240 0 0 0] Xor:[0 0 0 0]}
a-typ: *expr.Cmp
a-val: &{Op:1 Register:1 Data:[224 0 0 0]}
a-typ: *expr.Counter
a-val: &{Bytes:84 Packets:1}

```

can't found **expr.Masq**

stapelberg:
Can you provide a minimal program that illustrates the issue please?

PR #214, opened 2023-01-15
Masq marshal fix

turekt:
Hi,

there is a reported issue #213 which shows that expr.Masq is not unmarshaled as an expression by nftables lib. I have managed to reproduce this with a test case and added a fix.

Let me know what you think.


stapelberg (approved):

stapelberg:
Thank you!

junhwong:
Thank you!

Issue #171, opened 2022-06-07
[GetSets] Fetching Sets from table with maps fails


ziggie1984:
# Problem1

As soon as I want to fetch Sets of a table which has a map included the nflib reports an error:

`Error Fetching Sets -  could not determine data type 1cd`

When I delete the Map Fetching Sets works perfect.

# Problem 2

Fetching tables reports very high numbers for `Use    uint32 // NFTA_TABLE_USE (Number of chains in table)` for example:

Table has only 2 chains but GetTable reports very high number:

2022/06/07 09:41:28 &{filter 100663296 0 1} 

Is there maybe a short workaround to make it work?

ziggie1984:
FYI:

The problem lies here: https://github.com/google/nftables/blob/main/set.go#L705

In case the datatype is concatenated we need to call the `validateKeyType` Function which will fix the issue and check if all concatenated types are valid.
```
invalidMagic, ok := validateKeyType(nftMagic); !ok {
				return nil, fmt.Errorf("could not determine data type %+v", invalidMagic)
			} 
```

ziggie1984:
Will run some more tests maybe also include the possibility to decode a concatenated type (separate function), so that one can easily compare the data types before changing the set.

PR #217, opened 2023-03-13
Fix getting concatenated data types for maps

konradh:
This also implements parsing of concatenated data types and makes parsing datatypes of sets from magic values more efficient by using a map instead of looping over an array.

I think this fixes Problem 1 from https://github.com/google/nftables/issues/171.

### DIFF
--- a/set.go
+++ b/set.go
@@ -117,53 +117,53 @@ var (
 	TypeTimeDay     = SetDatatype{Name: "day", Bytes: 1, nftMagic: 45}
 	TypeCGroupV2    = SetDatatype{Name: "cgroupsv2", Bytes: 8, nftMagic: 46}
 
-	nftDatatypes = map[string]SetDatatype{
-		TypeVerdict.Name:     TypeVerdict,
-		TypeNFProto.Name:     TypeNFProto,
-		TypeBitmask.Name:     TypeBitmask,
-		TypeInteger.Name:     TypeInteger,
-		TypeString.Name:      TypeString,
-		TypeLLAddr.Name:      TypeLLAddr,
-		TypeIPAddr.Name:      TypeIPAddr,
-		TypeIP6Addr.Name:     TypeIP6Addr,
-		TypeEtherAddr.Name:   TypeEtherAddr,
-		TypeEtherType.Name:   TypeEtherType,
-		TypeARPOp.Name:       TypeARPOp,
-		TypeInetProto.Name:   TypeInetProto,
-		TypeInetService.Name: TypeInetService,
-		TypeICMPType.Name:    TypeICMPType,
-		TypeTCPFlag.Name:     TypeTCPFlag,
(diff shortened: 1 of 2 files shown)


stapelberg:
Looks reasonable to me, but I haven’t worked much with sets.

cc @sbezverk and @turekt in case they have any additional comments

Otherwise, I’ll merge this next week

set.go
```diff
@@ -678,60 +691,46 @@ func setsFromMsg(msg netlink.Message) (*Set, error) {
...
-		for _, dt := range nftDatatypes {
-			if t == dt.nftMagic {
-				found = true
-			}
-		}
-		if !found {
-			invalidTypes = append(invalidTypes, t)
-			valid = false
+func parseSetDatatype(magic uint32) (SetDatatype, error) {
+	types := make([]SetDatatype, 0, 32/SetConcatTypeBits)
```
turekt, line 717: This line made me wonder whether cap of 5 types is valid so I researched a bit and found this line in the original source https://git.netfilter.org/nftables/tree/src/datatype.c?id=1acc2fd48c755a8931fa87b8d0560b750316059f#n1165

Code never restricts concat type length to 5 and it is possible to add 6 types via the CLI:
```
$ nft flush ruleset
$ nft add table ip filter
$ nft add set filter setA { type nf_proto . ipv4_addr . inet_service . ipv4_addr . ipv6_addr . ipv6_addr \; flags timeout \; }
$ nft list ruleset
table ip filter {
	set setA {
		type nf_proto . ipv4_addr . inet_service . ipv4_addr . ipv6_addr . ipv6_addr
		flags timeout
	}
}
```

In case there is an overflow, the first type gets either ignored or converted (type is reduced to two bits, e.g. `nf_proto`), meaning that the cap is not specifically checked:
```
$ nft flush ruleset
$ nft add table ip filter
$ # ether_type is 10 so it gets reduced to two bits and converted into nf_proto (2)
$ nft add set filter setB { type ether_type . ipv4_addr . inet_service . ipv4_addr . ipv6_addr . ipv6_addr \; flags timeout \; }
$ nft list ruleset
table ip filter {
	set setB {
		type nf_proto . ipv4_addr . inet_service . ipv4_addr . ipv6_addr . ipv6_addr
		flags timeout
	}
}
```

Now, I am not sure whether using 6 types is a legitimate use case but I see that we check if there is more than 5 types beforehand https://github.com/google/nftables/blob/2729c5a5ee8e93e29b6fe0f857eda704158e19c9/set.go#L197 and with this check, the cap makes sense.

I will leave it up to @konradh to decide if he wants to leave this cap or not.

cc @stapelberg for awareness of this limitation, maybe good to note down in case someone raises an issue in the future.

turekt (approved):

set.go
stapelberg: @konradh what do you think?

PR #219, opened 2023-03-13
Drop unnecessary table rewrite in GetRules()

004helix:
Some old kernels (centos-7) send all rules via netlink ignoring table and chain names in GetRules()
There is no way to identify which table and chain contain the rule without this fix.

Moreover, there is no need to rewrite the table after commit 3e042f7


stapelberg:
Let me know when you have signed the CLA and I can review this PR. Thanks.

004helix:
It's already signed

stapelberg:
Thanks!

PR #221, opened 2023-04-15
Add libnftables-compatible TLV parser/serializer for rule user data

stv0g:
This TLV format is compatible with the one used by libnftables. Hence with this change, we can now de/encode comments and other user data information which is compatible with the Netfilter command line tooling.

I added tests for parsing/serializing strings and uint32 as well as an interoperability tests between the `nft` command line tool and this package.

### DIFF
--- a/userdata/userdata.go
+++ b/userdata/userdata.go
@@ -1,0 +1,94 @@
+// Copyright 2018 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package userdata implements a TLV parser/serializer for libnftables-compatible comments
+package userdata
+
+import (
+	"bytes"
+	"encoding/binary"
+)
+
+type Type byte
+
+// TLV type values are defined in:
+// https://git.netfilter.org/iptables/tree/iptables/nft.c?id=73611d5582e72367a698faf1b5301c836e981465#n1659
+const (
+	TypeComment Type = iota
+	TypeEbtablesPolicy
+
+	TypesCount
+)
+
+func Append(udata []byte, typ Type, data []byte) []byte {
+	udata = append(udata, byte(typ), byte(len(data)))
+	udata = append(udata, data...)
+
+	return udata
+}
+
+func Get(udata []byte, styp Type) []byte {
+	for {
+		if len(udata) < 2 {
+			break
+		}
+
+		typ := Type(udata[0])
+		length := int(udata[1])
+		data := udata[2 : 2+length]
+
+		if styp == typ {
+			return data
+		}
+
+		if len(udata) < 2+length {
+			break
+		} else {
+			udata = udata[2+length:]
+		}
+	}
+
+	return nil
+}
+
+func AppendUint32(udata []byte, typ Type, num uint32) []byte {
+	data := binary.LittleEndian.AppendUint32(nil, num)
+
+	return Append(udata, typ, data)
+}
+
+func GetUint32(udata []byte, typ Type) (uint32, bool) {
+	data := Get(udata, typ)
+	if data == nil {
+		return 0, false
+	}
+
+	return binary.LittleEndian.Uint32(data), true
+}
+
+func AppendString(udata []byte, typ Type, str string) []byte {
+	data := append([]byte(str), 0)
+	return Append(udata, typ, data)
+}
+
+func GetString(udata []byte, typ Type) (string, bool) {
+	data := Get(udata, typ)
+	if data == nil {
+		return "", false
+	}
+
+	data, _ = bytes.CutSuffix(data, []byte{0})
+
+	return string(data), true
+}
--- a/userdata/userdata_cli_interop_test.go
+++ b/userdata/userdata_cli_interop_test.go
@@ -1,0 +1,233 @@
+// Copyright 2018 Google LLC. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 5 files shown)


stv0g:
@stapelberg or @sbezverk would you mind to review/approve this PR?

userdata/userdata.go
```diff
@@ -0,0 +1,90 @@
+// Copyright 2023 Steffen Vogel.
```
stapelberg, line 1: Here and elsewhere: this needs to be Copyright Google LLC, not any individual’s name

userdata/userdata_cli_interop_test.go
```diff
@@ -0,0 +1,229 @@
...
+		Chain:    chain,
+		UserData: userdata.PutString(nil, userdata.TypeComment, wantComment),
+	})
+
+	if err := c.Flush(); err != nil {
+		t.Fatal(err)
+	}
+
+	out := bytes.NewBuffer(nil)
+	d := exec.Command("nft", "-j", "list", "table", "userdata-table")
```
stapelberg, line 108: Tests that modify system state should be guarded behind a flag, like enableSysTests in nftables_test.go.

Even better: make these tests use a throwaway network namespace so that they are safe to run without affecting the rest of the system.

userdata/userdata_cli_interop_test.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	if len(rules) != 1 {
+		t.Fatal()
+	}
+
+	if comment, ok := userdata.GetString(rules[0].UserData, userdata.TypeComment); !ok {
+		t.Fatalf("failed to find comment")
+	} else if comment != wantComment {
+		t.Fatalf("comment mismatch %s != %s", comment, wantComment)
```
stapelberg, line 227: Use %q instead of %s here so that special characters are easily recognizeable

userdata/userdata_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/google/nftables"
+	"github.com/google/nftables/userdata"
+)
+
+func TestUserDataComment(t *testing.T) {
+	r := nftables.Rule{}
+
+	wantComment := "this is my comment"
+	want := []byte{0, 19, 116, 104, 105, 115, 32, 105, 115, 32, 109, 121, 32, 99, 111, 109, 109, 101, 110, 116, 0}
```
stapelberg, line 30: Can you construct the human-readable part of the `want` byte slice by using []byte(wantComment) so that it’s clearer to see please?

userdata/userdata.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type Type byte
+
+const (
+	TypeComment Type = iota
+	TypeRuleID  Type = 100 // custom extension
+)
+
+func Put(udata []byte, typ Type, data []byte) []byte {
```
stapelberg, line 30: The semantics of this function are like the builtin `append`. I wonder if we should rename it from Put to Append?

userdata/userdata.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"bytes"
+	"encoding/binary"
+)
+
+type Type byte
+
+const (
+	TypeComment Type = iota
+	TypeRuleID  Type = 100 // custom extension
```
stapelberg, line 27: Where is this defined upstream? Can you add a link?

userdata/userdata.go
```diff
@@ -0,0 +1,90 @@
...
+			break
+		} else {
+			udata = udata[2+length:]
+		}
+	}
+
+	return nil
+}
+
+func PutInt(udata []byte, typ Type, num uint32) []byte {
```
stapelberg, line 61: Should this be PutUint32, or is it somehow clear that all ints must be uint32 and there can never be any other type?

userdata/userdata_test.go
```diff
@@ -0,0 +1,66 @@
...
+func TestInt(t *testing.T) {
+	r := nftables.Rule{}
+
+	wantRuleID := uint32(1234)
+	want := []byte{100, 4, 210, 4, 0, 0}
+
+	r.UserData = userdata.PutInt(r.UserData, userdata.TypeRuleID, wantRuleID)
+
+	if !bytes.Equal(r.UserData, want) {
+		t.Fatalf("UserData mismatch: %s != %s",
```
stapelberg, line 56: Would %x not work here (instead of calling hex.EncodeToString manually)?

stapelberg (changes requested):
Thanks! First review pass

userdata/userdata.go
stv0g: Mhh, okay. I disagree with this as in Germany, only individuals can be the owner of a copyrighted work.
And I already have signed a CLA with which I provide Google all rights to use my contribution.

Nevertheless, I changed it..

userdata/userdata_cli_interop_test.go
stv0g: Done. To reuse the code between `nftables_test.go` and `userdata_cli_interop_test.go`, I moved the setup/teardown of the NFT connection to `internal/nftest`.

userdata/userdata_cli_interop_test.go
stv0g: Done

userdata/userdata_test.go
stv0g: Done.

userdata/userdata.go
stv0g: I've renamed the `Put*()` methods to `Append*()`.

userdata/userdata.go
stv0g: I've added a link as well as a second type constant of which I was not aware yet.

userdata/userdata.go
stv0g: I agree. Changed it..

userdata/userdata_cli_interop_test.go
```diff
@@ -0,0 +1,229 @@
...
+	"bytes"
+	"encoding/json"
+	"os/exec"
+	"testing"
+
+	"github.com/google/nftables"
+	"github.com/google/nftables/userdata"
+)
+
+type NftCliMetainfo struct {
```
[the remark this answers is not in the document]
stv0g: Done.

userdata/userdata_test.go
stv0g: Thanks, I didnt knew `%x` yet :)

stv0g:
Hi @stapelberg,

thanks for the comments. I addressed all of them :-)

stapelberg (approved):

stapelberg:
Thank you!

PR #222, opened 2023-04-18
Implement socket expr

black-desk:


black-desk:
The socket expr seems to be the only way to use cgroupv2 with nftables.

expr/socket.go
```diff
@@ -0,0 +1,73 @@
+package expr
```
stapelberg, line 1: Please copy the license header from the other files to the top of this file.

stapelberg (changes requested):
Looks good overall. Can you add a test as well please?

Not sure what you mean regarding the cgroupv2 comment — how is this related? Can you explain in more detail?

Thanks.

black-desk:
Fine I will add tests later.

---

> Not sure what you mean regarding the cgroupv2 comment — how is this related? Can you explain in more detail? 
>
> Thanks.

This package can create cgroupv2 set:

https://github.com/google/nftables/blob/1510be9a554a67c6e5b34b552a5fc79f7f78c050/set.go#L118

But we cannot use cgroupv2 set for lookup before this PR, as the `meta` expr can only get cgroup v1 id:

https://wiki.nftables.org/wiki-nftables/index.php/Matching_packet_metainformation#Matching_by_security_selectors

the cgroupv1 id is an int32, while the socket cgroupv2 is an uint64: 

https://git.netfilter.org/nftables/tree/src/datatype.c#n1432

I do some search and I found this: https://www.spinics.net/lists/netfilter/msg60145.html

seems we can only get cgroupv2 by using `socket`

expr/socket.go
black-desk: I change 2018 to 2023.

expr/socket.go
```diff
@@ -0,0 +1,87 @@
...
+	SocketKeyWildcard      SocketKey = NFT_SOCKET_WILDCARD
+	SocketKeyCgroupv2      SocketKey = NFT_SOCKET_CGROUPV2
+)
+
+func (e *Socket) marshal(fam byte) ([]byte, error) {
+	exprData, err := netlink.MarshalAttributes(
+		[]netlink.Attribute{
+			{Type: NFTA_SOCKET_DREG, Data: binaryutil.BigEndian.PutUint32(e.Register)},
+			{Type: NFTA_SOCKET_KEY, Data: binaryutil.BigEndian.PutUint32(uint32(e.Key))},
+			{Type: NFTA_SOCKET_LEVEL, Data: binaryutil.BigEndian.PutUint32(uint32(e.Level))},
```
black-desk, line 56: `level` is only used when key == cgroup. But the nft utils written in c always encoding `level`. So I do this also.

stapelberg (changes requested):

expr/socket.go
black-desk: Done.

stapelberg (approved):

stapelberg:
Thanks!

Issue #177, opened 2022-08-28
Set with key type nftables.TypeIFName not working


rdmcguire:
Creating a set with Type: nftables.TypeIFName seems to work, but the set acts strangely.

*Sample code for set:*
```go
        conn.AddSet(&nftables.Set{
                Table:   table,
                Name:    "test_set",
                KeyType: nftables.TypeIFName,
        }, nil)
```

I also created a set called "manual_set" manually, and they both list:

*nftables pkg set and nft cli set*
```sh
% nft 'add set test_table manual_set { type ifname; }'
% nft list table test_table
table ip test_table {
        set test_set {
                type ifname
        }

        set manual_set {
                type ifname
        }
}
```

I then added an element into each set using nft cli and they both seem to succeed, however ***the set created by this package shows empty elements***.

*Added Elements*
```sh
% nft 'add element test_table test_set { "wg0" }'
% nft 'add element test_table manual_set { "wg0" }'
% nft list table test_table                                                                      ☸ kubernetes-admin@50w_k8s:argocd 
table ip test_table {
        set test_set {
                type ifname
                elements = { "" }
        }

        set manual_set {
                type ifname
                elements = { "wg0" }
        }
}
```

Additionally, trying to add an element through this package fails for either table, though I do wonder if I've missed something in my code:

*Attempt to add set element to both tables*
```go
        // github.com/google/nftables set and manual set
        testSet, _ := c.GetSetByName(table, "test_set")
        manualSet, _ := c.GetSetByName(table, "manual_set")

        // Second Element
        elements := []nftables.SetElement{{Key: []byte("wg1")}}

        c.SetAddElements(manualSet, elements)
        c.SetAddElements(testSet, elements)

        if err := c.Flush(); err != nil {
                log.Panicf("Error: %s", err)
        }
```

Running returns error: "conn.Receive: netlink receive: invalid argument"

I've tried a few different key types and have only had success with TypeFamilyIPv4.

turekt:
Hi @rdmcguire,

I have checked your issue and it seems that there is a bug in the set marshaling logic in the nftables go lib. If we observe the messages sent by the `nft` cmdline tool:
```
[
  // NFTA_SET_TABLE, "filter\x00"
  [{nla_len=11, nla_type=0x1}, "\x66\x69\x6c\x74\x65\x72\x00"], 
  // NFTA_SET_NAME, "test_set\x00"
[...]

Issue #225, opened 2023-05-15
AddSet IPv4 wrong byte order on Ubuntu 22.04


eldevel:
I am trying to add a set with ipv4 addresses, but the ip address bytes are reversed. Here is my code:
```
package main

import (
	"github.com/google/nftables"
	"github.com/vishvananda/netns"
	"net"
)

func main() {
	ns, _ := netns.Get()
	conn := nftables.Conn{NetNS: int(ns)}

	conn.FlushRuleset()

	mangle := conn.AddTable(&nftables.Table{
		Name:   "mangle",
		Family: nftables.TableFamilyINet,
	})

	testSet := &nftables.Set{
		Name:    "test_v4",
		Table:   mangle,
		KeyType: nftables.TypeIPAddr,
	}

	conn.AddSet(testSet, []nftables.SetElement{{
		Key: []byte(net.ParseIP("192.168.1.64").To4()),
	}})

	conn.Flush()
}
```

But the result is:
```
# nft list ruleset
table inet mangle {
        set test_v4 {
                type ipv4_addr
                elements = { 64.1.168.192 }
        }
}
```

Some additional information:

```
# go version
go version go1.18.1 linux/amd64

# nft -v
nftables v1.0.2 (Lester Gooch)

# uname -r -v
5.15.0-71-generic #78-Ubuntu SMP Tue Apr 18 09:00:29 UTC 2023

# cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.2 LTS

# lscpu | grep Order
Byte Order:                      Little Endian
```

DamienDelporte:
+1 Was about to open the same exact issue :)
```
$ go version
go version go1.19 linux/amd64

$ nft -v
nftables v0.9.3 (Topsy)

$ uname -r -v
5.14.0-1059-oem #67-Ubuntu SMP Mon Mar 13 14:22:10 UTC 2023

$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.6 LTS"

$ lscpu | grep Order
Byte Order:                      Little Endian
```

stapelberg:
[...]

PR #226, opened 2023-05-16
Implement set KeyByteOrder

turekt:
Hi,

this PR reverts PR #180 and introduces `KeyByteOrder` field to sets to allow lib users to specify key byte order. PR #180 was a change that introduced a default userdata TLV that specifies key endianess and was implemented to fix issue #177 and conform to main nftables C implementation. This obviously broke setting IPs to sets and there is no other way for users to specify which byte order should be used. Therefore, this change now allows specifying key byte order and should be backwards compatible since not specifying `KeyByteOrder` will marshal bytes as it would prior to PR #180.

With this, issue #225 should be fixed. Issue #177 remains, but can now be fixed by specifying `s.KeyByteOrder = binaryutil.NativeEndian`.

Let me know what you think.

### DIFF
--- a/set.go
+++ b/set.go
@@ -261,6 +261,9 @@ type Set struct {
 	Timeout       time.Duration
 	KeyType       SetDatatype
 	DataType      SetDatatype
+	// Either host (binaryutil.NativeEndian) or big (binaryutil.BigEndian) endian as per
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks for the fix!

Issue #224, opened 2023-04-26
Feature: add support for monitor


singchia:
It seems this library doesn't support MONITOR, which allows to listen to netlink events like creation and deletion of objects.

For example: 

`nft monitor chains` to monitor changes on chains.

If so, I would like to submit a PR to support this.

singchia:
OK, working on it.

cheina97:
This feature is still a work in progress?

singchia:
Yes, new PR will be submitted recently.

cheina97:
Thanks @singchia ❤️

singchia:
the PR is here #250 
 hope we can make it faster to be merged.

PR #228, opened 2023-05-17
Feat: add nftables objects monitor function and unit tests

singchia:
Here it comes for the issue #224


singchia:
I noticed that other PR has conflicts with the feature, so I'll resolve them, split the commits and re-PR with more description.

PR #230, opened 2023-05-20
Parse set element counters

jronak:
This PR adds the ability to parse counter data of the set elements. 
When a set is created with `counter` set, each set element maintains a counter that is incremented when the element is accessed. 

Today, you can access the set elements counter like this:
```
> nft list set my-table my-set
table ip my-table {
	set my-set {
		typeof ip saddr
		counter
		elements = { 10.0.0.1 counter packets 0 bytes 0, 10.0.0.2 counter packets 142833 bytes 422829523 }
	}
}
```

I was able to test the changes locally, but there is no easy way to add tests as a counter cannot be inserted in the set element explicitly.

### DIFF
--- a/set.go
+++ b/set.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/google/nftables/expr"
+	"github.com/google/nftables/internal/parseexprfunc"
 
 	"github.com/google/nftables/binaryutil"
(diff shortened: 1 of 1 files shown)


stapelberg (approved):

stapelberg:
Thanks!

PR #231, opened 2023-05-31
Add address field for expr tproxy

black-desk:


black-desk:
@MrRedhat please sign your CLA.

expr/tproxy.go
```diff
@@ -25,22 +25,36 @@ import (
...
 	// NFTA_TPROXY_FAMILY defines attribute for a table family
 	NFTA_TPROXY_FAMILY = 0x01
-	// NFTA_TPROXY_REG defines attribute for a register carrying redirection port value
-	NFTA_TPROXY_REG = 0x03
+	// NFTA_TPROXY_REG_ADDR defines attribute for a register carrying redirection address value
+	NFTA_TPROXY_REG_ADDR = 0x02
+	// NFTA_TPROXY_REG_PORT defines attribute for a register carrying redirection port value
+	NFTA_TPROXY_REG_PORT = 0x03
+
+	NFTA_TPROXY_REG = NFTA_TPROXY_REG_PORT // DEPRECATED
```
stapelberg, line 33: Why is this constant marked as deprecated? Could we just remove it instead? (I don’t see any usages on github.com/search)

The format for deprecations in Go is:
```go
// Deprecated: use NFTA_TPROXY_REG_PORT instead
```

black-desk: I cannot found NFTA_TPROXY_REG in the `nft` git repo,
so I update it to NFTA_TPROXY_REG_PORT to match `nft`'s define:

https://git.netfilter.org/nftables/tree/include/linux/netfilter/nf_tables.h?id=25e7b99cc450490c38becb03d8bddd0199cfd3f9#n1464

This constant is public. I think that we should not remove it, as removing it seems to be a breaking change?

> I don’t see any usages on github.com/search

Fine. Let's remove it.

stapelberg (approved):

PR #234, opened 2023-08-02
hash: Fix hash seed conditional

methril:
Fix the marshall of the hash seed to be conditional, only if it is explicitly set, we need to add it to the kernel as stated on the libnftl and nftables projects.
Refence: https://git.netfilter.org/nftables/tree/src/netlink_linearize.c#n174

Otherwise, having a hash expression similar to this:
```
ip daddr set jhash tcp sport mod 2 seed 0x0 map { 0 : 192.168.0.1, 1 : 192.168.2.2 }
```
end up setting only the first IP and ignoring the second one.

### DIFF
--- a/expr/hash.go
+++ b/expr/hash.go
@@ -41,15 +41,22 @@ type Hash struct {
 }
 
 func (e *Hash) marshal(fam byte) ([]byte, error) {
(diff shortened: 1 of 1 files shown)


stapelberg:
Thanks!

Issue #238, opened 2023-08-16
Named quotas and their usage in map


ProjectIcarusX:
Hi,

Is it possible to add/create/list/delete named quotas through the library?
Also is it possible to use them as the map values? 
NFTables support them and they can be added through CLI. The library currently lists maps with quota values as normal sets with value undefined.

PR #244, opened 2023-10-22
Support for quota as object

turekt:
Hi,

I have added support for quota objects as stated in #238. 

One thing to note: I have implemented this object type in a way that aligns with the existing implementation approach of counter object that is already implemented in this lib. It might be good to reconsider the existing implementation approach since object data can be specified as references to existing expression structs implemented in `expr` package and the expression data marshal logic could be reused. At the moment this requires careful rewrite of larger parts of code (`expr.Any` interface where marshaling of the object should be separated from marshaling of the complete expression and `nftables.Obj` interface implementation at the very least) and seems a bit complex since it easily breaks backwards compatibility. Current implementation should work for the time being.

Let me know what you think.

### DIFF
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -6077,6 +6077,198 @@ func TestGetRulesObjref(t *testing.T) {
 	}
 }
 
+func TestAddQuotaObj(t *testing.T) {
+	conn, newNS := nftest.OpenSystemConn(t, *enableSysTests)
+	defer nftest.CleanupSystemConn(t, newNS)
+	conn.FlushRuleset()
+	defer conn.FlushRuleset()
+
+	table := &nftables.Table{
+		Name:   "quota_demo",
+		Family: nftables.TableFamilyIPv4,
+	}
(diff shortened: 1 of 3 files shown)


stapelberg:
> One thing to note: I have implemented this object type in a way that aligns with the existing implementation approach of counter object that is already implemented in this lib. It might be good to reconsider the existing implementation approach since object data can be specified as references to existing expression structs implemented in `expr` package and the expression data marshal logic could be reused. At the moment this requires careful rewrite of larger parts of code (`expr.Any` interface where marshaling of the object should be separated from marshaling of the complete expression and `nftables.Obj` interface implementation at the very least) and seems a bit complex since it easily breaks backwards compatibility. Current implementation should work for the time being.

Could you file a separate issue for this refactoring idea, perhaps with a bit of draft code to illustrate what it would look like? Thanks :)

stapelberg:
Hey @turekt could you contact me via email for a question please? Email michael AT stapelberg DOT ch; thank you :)

Issue #245, opened 2023-10-24
How to get an error when try to add an existed table?


black-desk:
Conn.AddTable will not return any error when I try to add a table already existed.

Note:

`nft create table ...` will report an error when I try to do so, but `nft add table ...` not.

PR #246, opened 2023-10-24
feat: add Conn.CreateTable

black-desk:
`Conn.AddTable` use netlink.Create which will not emit an error
if the table we want to create already existed,
just like the `nft add table ...` command works.

The caller should use netlink.Excl to
get an EEXIST error for that already existed,

So I add another method `Conn.CreateTable`
which works just like `nft create table ...` command.

Related: #245
Signed-off-by: black-desk <me@black-desk.cn>

### DIFF
--- a/table.go
+++ b/table.go
@@ -63,9 +63,7 @@ func (cc *Conn) DelTable(t *Table) {
 	})
 }
 
-// AddTable adds the specified Table. See also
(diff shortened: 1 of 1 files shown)


black-desk:
@stapelberg Maybe we need more CreateXXX methods.

stapelberg (approved):

PR #252, opened 2024-01-04
NAT: prefix test

cheina97:
This PR adds unit tests for PR #251

In particular it adds a unit test for this rule `nft add rule nat prerouting ip daddr 10.0.0.0/24 dnat prefix to 20.0.0.0/24`

The PR includes some utilities to get the first and last IP of a subnet (they include unit tests)


util.go
```diff
@@ -44,3 +45,35 @@ func (genmsg *NFGenMsg) Decode(b []byte) {
...
+	_, subnet, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return nil, err
+	}
+
+	mask := binary.BigEndian.Uint32(subnet.Mask)
+	ip := binary.BigEndian.Uint32(subnet.IP)
+
+	// find the final address
+	firstIP := make(net.IP, 4)
```
stapelberg, line 60: This is IPv4 specific. Can the function also be made to work with IPv6? I know NAT is rare with IPv6, but sometimes used.

util.go
```diff
@@ -44,3 +45,35 @@ func (genmsg *NFGenMsg) Decode(b []byte) {
...
+func GetLastIPFromCIDR(cidr string) (*net.IP, error) {
+	_, subnet, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return nil, err
+	}
+	mask := binary.BigEndian.Uint32(subnet.Mask)
+	ip := binary.BigEndian.Uint32(subnet.IP)
+	// find the final address
+	lastIP := make(net.IP, 4)
+	binary.BigEndian.PutUint32(lastIP, (ip&mask)|(mask^0xffffffff))
```
stapelberg, line 76: What does the bitwise operation do? Deserves a comment.

stapelberg (changes requested):
Thanks for the PR!

cheina97:
Hi @stapelberg, I've left a commit with the fixes you asked.
If it is ok I can squash the commits.

util.go
stapelberg: This doesn’t seem to be resolved? I don’t see a comment explaining the bitwise operation.

cheina97: I'm sorry, I've commented on another bitwise operation in nftables_test.go. Now I've added the comment here. Let me know if you think it is clear.

stapelberg:
Thanks! Merged as commit https://github.com/google/nftables/commit/33ee8df9d8e2cbe47a998c97fe6b842cf00b324a with slight adjustments in naming/documentation.

PR #256, opened 2024-03-15
Replace TODO const by golang.org/x/sys/unix

LionelJouin:
Replace some hardcoded consts by the ones provided in golang.org/x/sys/unix


stapelberg (approved):

stapelberg:
Thanks for the cleanup!

PR #258, opened 2024-04-13
List single table or chain by name

turekt:
Hi,

I think that current nftables package is missing the `nft list chain` and `nft list table` functionality. Without this functionality, current user code that wants to manage one specific table or chain has to fetch all of tables/chains and then iterate until it matches the desired table/chain by name (and family). This pull request adds the ability to fetch the table/chain with a single request that returns the desired table/chain immediately.

Current change should not affect existing usage of `ListTablesOfFamily` or `ListChainsOfTableFamily`. Tests are also added.

Let me know what you think.

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -193,6 +193,43 @@ func (cc *Conn) ListChains() ([]*Chain, error) {
 	return cc.ListChainsOfTableFamily(TableFamilyUnspecified)
 }
 
+// ListChain returns a single chain configured in the specified table
+func (cc *Conn) ListChain(table *Table, chain string) (*Chain, error) {
+	conn, closer, err := cc.netlinkConn()
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = closer() }()
+
+	attrs := []netlink.Attribute{
+		{Type: unix.NFTA_TABLE_NAME, Data: []byte(table.Name + "\x00")},
+		{Type: unix.NFTA_CHAIN_NAME, Data: []byte(chain + "\x00")},
+	}
+	msg := netlink.Message{
+		Header: netlink.Header{
+			Type:  netlink.HeaderType((unix.NFNL_SUBSYS_NFTABLES << 8) | unix.NFT_MSG_GETCHAIN),
+			Flags: netlink.Request,
+		},
+		Data: append(extraHeader(uint8(table.Family), 0), cc.marshalAttr(attrs)...),
+	}
+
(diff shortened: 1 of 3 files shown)


chain.go
```diff
@@ -193,6 +193,62 @@ func (cc *Conn) ListChains() ([]*Chain, error) {
...
+		return nil, err
+	}
+	defer func() { _ = closer() }()
+
+	tableAttr := []netlink.Attribute{
+		{Type: unix.NFTA_TABLE_NAME, Data: []byte(table.Name + "\x00")},
+	}
+	flags := netlink.Request
+	msgs := []netlink.Message{
+		{
```
stapelberg, line 208: The result of this first GETTABLE request seems to be discarded later? Could you add a comment explaining why two requests need to be sent instead of just the GETCHAIN request?

chain.go
```diff
@@ -193,6 +193,62 @@ func (cc *Conn) ListChains() ([]*Chain, error) {
...
+				Flags: flags,
+			},
+			Data: append(extraHeader(uint8(table.Family), 0), cc.marshalAttr(tableAttr)...),
+		},
+		{
+			Header: netlink.Header{
+				Type:  netlink.HeaderType((unix.NFNL_SUBSYS_NFTABLES << 8) | unix.NFT_MSG_GETCHAIN),
+				Flags: flags | netlink.Acknowledge,
+			},
+			Data: append(extraHeader(uint8(table.Family), 0), cc.marshalAttr(append(tableAttr, netlink.Attribute{Type: unix.NFTA_CHAIN_NAME, Data: []byte(chain + "\x00")}))...),
```
stapelberg, line 220: Assign the cc.marshalAttr() to a helper variable, this line is getting rather long…

stapelberg (commented):
Seems reasonable overall

chain.go
turekt: Thanks for pointing this out. It seems that you were right. The `nft` CLI, which I used as reference, seems to send a message with type 2561 (GET_TABLE) before sending type 2564 (GET_CHAIN). I took a look at libnftnl source code and confirmed that a single GET_CHAIN request is enough: http://git.netfilter.org/libnftnl/tree/examples/nft-chain-get.c?id=be0bae0ad31b0adb506f96de083f52a2bd0d4fbf#n89

chain.go
turekt: Done.

stapelberg (approved):

Issue #253, opened 2024-01-29
Objects implementation refactor


turekt:
Hi,

I am creating an issue as discussed in PR #244.

# TL;DR

Current object implementation seems to consist of mostly duplicated code from the `expr` package. This issue tries to present the problem and propose an implementation which reuses the `expr` implementation, removes duplicated code and provides immediate support of all stateful object types supported by nftables.

# Observations

During implementation of quota objects I noticed that the elements in the object (`Obj`) structs are duplicated. Observe the following counter object `CounterObj` in `counter.go` against `Counter` in `expr/counter.go`:
```go
type CounterObj struct {
	Table *Table
	Name  string // e.g. “fwded”

	Bytes   uint64
	Packets uint64
}
...
type Counter struct {
	Bytes   uint64
	Packets uint64
}
```

Note how expression elements in the struct (Packets and Bytes) are duplicated. Here is their corresponding `unmarshal`ing logic:
```go
func (c *CounterObj) unmarshal(ad *netlink.AttributeDecoder) error {
	for ad.Next() {
		switch ad.Type() {
		case unix.NFTA_COUNTER_BYTES:
			c.Bytes = ad.Uint64()
		case unix.NFTA_COUNTER_PACKETS:
			c.Packets = ad.Uint64()
		}
	}
	return ad.Err()
}
...
func (e *Counter) unmarshal(fam byte, data []byte) error {
	ad, err := netlink.NewAttributeDecoder(data)
	if err != nil {
		return err
	}
	ad.ByteOrder = binary.BigEndian
	for ad.Next() {
		switch ad.Type() {
		case unix.NFTA_COUNTER_BYTES:
			e.Bytes = ad.Uint64()
		case unix.NFTA_COUNTER_PACKETS:
			e.Packets = ad.Uint64()
		}
	}
	return ad.Err()
}
```

Marshaling logic is a bit more different since object marshaling creates a complete expression, but the data specific implementation is duplicated:
```go
// returns only expression
func (e *Counter) marshal(fam byte) ([]byte, error) {
	data, err := netlink.MarshalAttributes([]netlink.Attribute{
		{Type: unix.NFTA_COUNTER_BYTES, Data: binaryutil.BigEndian.PutUint64(e.Bytes)},
		{Type: unix.NFTA_COUNTER_PACKETS, Data: binaryutil.BigEndian.PutUint64(e.Packets)},
	})
	if err != nil {
		return nil, err
	}

[...]

PR #259, opened 2024-04-17
Objects implementation refactor

turekt:
Hi,

as per our discussion in #253, I have implemented a change for a more generic implementation of nftables objects. In addition to the refactor I have added a test to demonstrate that already implemented expressions are now supported as objects and made a few lint changes. Note that some of the expressions are currently not implemented in the lib (i.e. synproxy, ct_helper, ct_expect, ...). These can now be easily added to the `expr` package to offer support. I'll see to adding them after we agree on this refactor change.

Let me know what you think.

### DIFF
--- a/counter.go
+++ b/counter.go
@@ -15,12 +15,11 @@
 package nftables
 
 import (
-	"github.com/google/nftables/binaryutil"
+	"github.com/google/nftables/expr"
 	"github.com/mdlayher/netlink"
 	"golang.org/x/sys/unix"
 )
 
-// CounterObj implements Obj.
 type CounterObj struct {
 	Table *Table
 	Name  string // e.g. “fwded”
@@ -41,29 +40,24 @@ func (c *CounterObj) unmarshal(ad *netlink.AttributeDecoder) error {
 	return ad.Err()
 }
 
+func (c *CounterObj) data() expr.Any {
+	return &expr.Counter{
+		Bytes:   c.Bytes,
+		Packets: c.Packets,
+	}
+}
+
+func (c *CounterObj) name() string {
+	return c.Name
+}
+func (c *CounterObj) objType() ObjType {
+	return ObjTypeCounter
+}
+
 func (c *CounterObj) table() *Table {
 	return c.Table
 }
 
 func (c *CounterObj) family() TableFamily {
 	return c.Table.Family
-}
-
-func (c *CounterObj) marshal(data bool) ([]byte, error) {
-	obj, err := netlink.MarshalAttributes([]netlink.Attribute{
-		{Type: unix.NFTA_COUNTER_BYTES, Data: binaryutil.BigEndian.PutUint64(c.Bytes)},
-		{Type: unix.NFTA_COUNTER_PACKETS, Data: binaryutil.BigEndian.PutUint64(c.Packets)},
-	})
-	if err != nil {
-		return nil, err
-	}
-	attrs := []netlink.Attribute{
-		{Type: unix.NFTA_OBJ_TABLE, Data: []byte(c.Table.Name + "\x00")},
-		{Type: unix.NFTA_OBJ_NAME, Data: []byte(c.Name + "\x00")},
-		{Type: unix.NFTA_OBJ_TYPE, Data: binaryutil.BigEndian.PutUint32(unix.NFT_OBJECT_COUNTER)},
-	}
-	if data {
-		attrs = append(attrs, netlink.Attribute{Type: unix.NLA_F_NESTED | unix.NFTA_OBJ_DATA, Data: obj})
-	}
-	return netlink.MarshalAttributes(attrs)
 }
--- a/obj.go
+++ b/obj.go
@@ -18,6 +18,9 @@ import (
 	"encoding/binary"
 	"fmt"
 
+	"github.com/google/nftables/binaryutil"
+	"github.com/google/nftables/expr"
+	"github.com/google/nftables/internal/parseexprfunc"
 	"github.com/mdlayher/netlink"
 	"golang.org/x/sys/unix"
 )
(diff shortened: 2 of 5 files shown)


counter.go
```diff
@@ -16,11 +16,12 @@ package nftables
 
 import (
 	"github.com/google/nftables/binaryutil"
+	"github.com/google/nftables/expr"
 	"github.com/mdlayher/netlink"
 	"golang.org/x/sys/unix"
 )
 
-// CounterObj implements Obj.
+// Deprecated: Use ObjAttr instead
```
stapelberg, line 24: Instead of deprecating these types, would it be possible to convert to ObjAttr under the covers? Maybe I’m not fully understanding the before/after, though. Would you mind showing an example change that users of CounterObj need to do?

expr/bitwise.go
```diff
@@ -54,6 +54,9 @@ func (e *Bitwise) marshal(fam byte) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	if dataOnly {
```
stapelberg, line 57: Instead of adding a boolean to all marshal functions, could we split this logic out into two separate functions? One to prepare the data (and return it), another one which calls the data preparation func and then the netlink.MarshalAttributes?

set.go
```diff
@@ -832,7 +832,7 @@ func (cc *Conn) GetSets(t *Table) ([]*Set, error) {
 
 	reply, err := receiveAckAware(conn, message.Header.Flags)
 	if err != nil {
-		return nil, fmt.Errorf("Receive: %v", err)
+		return nil, fmt.Errorf("receiveAckAware: %v", err)
```
stapelberg, line 835: Would you mind sending these unrelated fixes as a separate PR? I’d like to keep the refactoring as self-contained and minimal as possible.

stapelberg (commented):
Thanks, it’s very helpful to see the entire change laid out like that. I left a few detail questions.

set.go
turekt: Done. Will be addressed separately.

expr/bitwise.go
turekt: Agreed, refactored to separate functions approach.

counter.go
turekt: I've already changed CounterObj and QuoteObj to implement Obj interface as a form of compatibility with functions in `obj.go`. With this change, the only thing that users of CounterObj should change is the conversion of the returned object that implements the Obj interface.

Here are a few examples:
```go
// This will work since CounterObj implements Obj
obj, err := c.ResetObject(&nftables.CounterObj{
	Table: filter,
	Name:  "fwded",
})

// This will work as well
counter1 := c.AddObj(&nftables.CounterObj{
	Table:   table,
	Name:    "fwded1",
	Bytes:   1,
	Packets: 1,
})
// Also works, although counter1 is technically ObjAttr
obj1, err := c.GetObject(counter1)
// This wouldn't work anymore, obj1 is ObjAttr
rcounter1, ok := obj1.(*nftables.CounterObj)
// Previous line should be changed to this
rcounter1, ok := obj1.(*nftables.ObjAttr)
```

We could add a special case when we unmarshal ObjTypeCounter and ObjTypeQuota to return CounterObj and QuotaObj instead of ObjAttr, but this approach kind of beats the purpose of using ObjAttr for these two types since we would always leverage legacy types instead of new ObjAttr.

Let me know how to approach this.

turekt: Hi @stapelberg, 

this discussion has gone stale a bit. I have refactored this feature to make it backwards compatible. The implementation now detects when user passes the "new" `ObjAttr` type and then works with and returns the `ObjAttr` value, otherwise we default to "old" `QuotaObj` or `CounterObj`. I have implemented tests using both variants to demonstrate how it works.

Let me know what you think.

expr/expr.go
```diff
@@ -59,15 +59,31 @@ func Marshal(fam byte, e Any) ([]byte, error) {
...
 func Unmarshal(fam byte, data []byte, e Any) error {
 	return e.unmarshal(fam, data)
 }
 
 // exprsFromBytes parses nested raw expressions bytes
 // to construct nftables expressions
-func exprsFromBytes(fam byte, ad *netlink.AttributeDecoder, b []byte) ([]Any, error) {
+func exprsFromBytes(fam byte, ad *netlink.AttributeDecoder, args ...string) ([]Any, error) {
 	var exprs []Any
+	if len(args) > 0 {
```
stapelberg, line 75: Couldn’t this code path go into a separate function? It seems like this function now does two separate things depending on how it’s called.

obj.go
```diff
@@ -27,13 +30,70 @@ var (
...
 	table() *Table
 	family() TableFamily
-	unmarshal(*netlink.AttributeDecoder) error
-	marshal(data bool) ([]byte, error)
+	data() expr.Any
+	name() string
+	objType() ObjType
+}
+
+type ObjAttr struct {
```
stapelberg, line 72: Can you add a comment to explain what this type does on a logical level?

obj.go
```diff
@@ -150,6 +219,29 @@ func objFromMsg(msg netlink.Message) (Obj, error) {
 		case unix.NFTA_OBJ_TYPE:
 			objectType = ad.Uint32()
 		case unix.NFTA_OBJ_DATA:
+			if !returnLegacyType {
```
stapelberg, line 222: I think it would be a little cleaner if we reversed control flow here: the legacy type handling could go into a separate function, and this non-legacy code path should be the default (not indented).

obj.go
```diff
@@ -86,13 +155,13 @@ func (cc *Conn) DeleteObject(o Obj) {
 // GetObj is a legacy method that return all Obj that belongs
 // to the same table as the given one
```
stapelberg, line 156: Can you add a brief comment to explain the different behavior regarding legacy/non-legacy Obj types please?

stapelberg (commented):
Thanks, I think this looks good overall. Would it be possible to move the refactoring of the expr/*.go files into a separate commit? We could probably even submit that independently, as it should be a no-op, yes?

expr/expr.go
turekt: This change was implemented as part of #265 since it belongs to refactoring of the expr/*.go files.

stapelberg:
Alright, PR #265 is now merged, so you can rebase this one.

stapelberg (changes requested):

quota.go
```diff
@@ -63,7 +59,7 @@ func (q *QuotaObj) marshal(data bool) ([]byte, error) {
 	attrs := []netlink.Attribute{
 		{Type: unix.NFTA_OBJ_TABLE, Data: []byte(q.Table.Name + "\x00")},
 		{Type: unix.NFTA_OBJ_NAME, Data: []byte(q.Name + "\x00")},
-		{Type: unix.NFTA_OBJ_TYPE, Data: binaryutil.BigEndian.PutUint32(NFT_OBJECT_QUOTA)},
+		{Type: unix.NFTA_OBJ_TYPE, Data: binaryutil.BigEndian.PutUint32(unix.NFT_OBJECT_QUOTA)},
```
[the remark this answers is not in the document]
turekt: I believe that is correct. In all cases we marshal using new `NamedObj` and afterwards convert the `NamedObj` to appropriate legacy type if required.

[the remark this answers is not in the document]
stapelberg: Then why not remove the now dead code? :)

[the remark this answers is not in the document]
turekt: Fair enough.

stapelberg (approved):

PR #260, opened 2024-04-17
feat: add xt.Comment

thediveo:
This adds support for xt.Comment (including unit tests), the fixed length comment data elements in the iptables translation layer, as per https://elixir.bootlin.com/linux/v6.8.7/source/include/uapi/linux/netfilter/xt_comment.h#L7.

### DIFF
--- a/xt/comment.go
+++ b/xt/comment.go
@@ -1,0 +1,34 @@
+package xt
+
+import (
+	"bytes"
+	"fmt"
+)
+
+// CommentSize is the fixed size of a comment info xt blob, see:
+// https://elixir.bootlin.com/linux/v6.8.7/source/include/uapi/linux/netfilter/xt_comment.h#L5
+const CommentSize = 256
+
+// Comment gets marshalled and unmarshalled as a fixed-sized char array, filled
+// with zeros as necessary, see:
+// https://elixir.bootlin.com/linux/v6.8.7/source/include/uapi/linux/netfilter/xt_comment.h#L7
+type Comment string
+
+func (c *Comment) marshal(fam TableFamily, rev uint32) ([]byte, error) {
+	if len(*c) >= CommentSize {
+		return nil, fmt.Errorf("comment must be less than %d bytes, got %d bytes",
+			CommentSize, len(*c))
+	}
+	data := make([]byte, CommentSize)
+	copy(data, []byte(*c))
+	return data, nil
+}
+
+func (c *Comment) unmarshal(fam TableFamily, rev uint32, data []byte) error {
+	if len(data) != CommentSize {
+		return fmt.Errorf("malformed comment: got %d bytes, expected exactly %d bytes",
+			len(data), CommentSize)
+	}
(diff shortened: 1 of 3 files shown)


xt/comment.go
```diff
@@ -0,0 +1,31 @@
...
+const CommentSize = 256
+
+// Comment gets marshalled and unmarshalled as a fixed-sized char array, filled
+// with zeros as necessary, see:
+// https://elixir.bootlin.com/linux/v6.8.7/source/include/uapi/linux/netfilter/xt_comment.h#L7
+type Comment string
+
+func (c *Comment) marshal(fam TableFamily, rev uint32) ([]byte, error) {
+	if len(*c) >= CommentSize {
+		return nil, fmt.Errorf("Comment must be less than %d bytes", CommentSize)
```
stapelberg, line 18: nit: here and below: start fmt.Errorf() args with a lowercase letter (“comment” in this case)

stapelberg, line 18: Please include the actual size of the comment as well for convenience of the reader of this error

xt/comment.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	data := make([]byte, CommentSize)
+	copy(data, []byte(*c))
+	return data, nil
+}
+
+func (c *Comment) unmarshal(fam TableFamily, rev uint32, data []byte) error {
+	if len(data) != CommentSize {
+		return fmt.Errorf("Comment takes exactly %d bytes", CommentSize)
```
stapelberg, line 27: same here. How about:

return fmt.Errorf("malformed comment: got %d bytes, expected exactly %d bytes", len(data), CommentSize)

xt/comment_test.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+			if !reflect.DeepEqual(tt.info, recoveredInfo) {
+				t.Fatalf("original %+v and recovered %+v are different", tt.info, recoveredInfo)
+			}
+		})
+	}
+
+	oversizeddata := []byte(oversized)
+	var comment Comment
+	if err := (&comment).unmarshal(0, 0, oversizeddata); err == nil || err.Error() != fmt.Sprintf("Comment takes exactly %d bytes", CommentSize) {
```
stapelberg, line 60: Don’t duplicate the exact error message here, it’s annoying when changing the message :) 

Just check that an error was returned.

stapelberg (changes requested):
Thanks for your PR!

xt/comment.go
thediveo: I had assumed we would use the type identifiers here? (That's what we seem to use throughout this package, if I'm not mistaken)

xt/comment_test.go
thediveo: ...true, otoh I was trying to make sure that I don't get a different error that we then mistake. Easy to change.

stapelberg (approved):

thediveo:
Many thanks!

PR #271, opened 2024-08-30
set: add set support auto-merge

Asutorufa:


set.go
```diff
@@ -580,15 +581,27 @@ func (cc *Conn) AddSet(s *Set, vals []SetElement) error {
...
 	} else if s.KeyByteOrder == binaryutil.NativeEndian {
 		// Per https://git.netfilter.org/nftables/tree/src/mnl.c?id=187c6d01d35722618c2711bbc49262c286472c8f#n1165
-		tableInfo = append(tableInfo,
-			netlink.Attribute{Type: unix.NFTA_SET_USERDATA, Data: []byte("\x00\x04\x01\x00\x00\x00")})
+		userData = userdata.AppendUint32(userData, 0, 1)
 	}
+
+	if s.Interval && s.AutoMerge {
+		// https://git.netfilter.org/nftables/tree/src/mnl.c?id=187c6d01d35722618c2711bbc49262c286472c8f#n1174
+		userData = userdata.AppendUint32(userData, 2, 1)
```
stapelberg, line 598: The `2` in this line should be `NFTNL_UDATA_SET_MERGE_ELEMENTS`. Please introduce these constants from https://git.netfilter.org/libnftnl/tree/include/libnftnl/udata.h#n39

Same above (for `NFTNL_UDATA_SET_KEYBYTEORDER`) and below

stapelberg (changes requested):
Thanks for your PR!

set.go
Asutorufa: Thanks for your review, I introduced these constants at userdata package.

stapelberg (approved):

PR #272, opened 2024-09-01
Add ct expect support

turekt:
Continuation of NamedObj implementation - ct expect stateful object.

### DIFF
--- a/expr/ct.go
+++ b/expr/ct.go
@@ -194,3 +194,74 @@ func (c *CtHelper) unmarshal(fam byte, data []byte) error {
 	}
 	return ad.Err()
 }
+
+// From https://git.netfilter.org/libnftnl/tree/include/linux/netfilter/nf_tables.h?id=be0bae0ad31b0adb506f96de083f52a2bd0d4fbf#n1601
+// Currently not available in sys/unix
+const (
+	NFTA_CT_EXPECT_L3PROTO = 0x01
+	NFTA_CT_EXPECT_L4PROTO = 0x02
+	NFTA_CT_EXPECT_DPORT   = 0x03
+	NFTA_CT_EXPECT_TIMEOUT = 0x04
+	NFTA_CT_EXPECT_SIZE    = 0x05
+)
+
+type CtExpect struct {
+	L3Proto uint16
+	L4Proto uint8
+	DPort   uint16
+	Timeout uint32
+	Size    uint8
+}
+
+func (c *CtExpect) marshal(fam byte) ([]byte, error) {
+	exprData, err := c.marshalData(fam)
+	if err != nil {
+		return nil, err
+	}
+
+	return netlink.MarshalAttributes([]netlink.Attribute{
+		{Type: unix.NFTA_EXPR_NAME, Data: []byte("ctexpect\x00")},
+		{Type: unix.NLA_F_NESTED | unix.NFTA_EXPR_DATA, Data: exprData},
+	})
+}
+
+func (c *CtExpect) marshalData(fam byte) ([]byte, error) {
+	// all elements except l3proto must be defined
+	// per https://git.netfilter.org/nftables/tree/doc/stateful-objects.txt?id=db70959a5ccf2952b218f51c3d529e186a5a43bb#n119
+	// from man page: l3proto is derived from the table family by default
+	exprData := []netlink.Attribute{
+		{Type: NFTA_CT_EXPECT_L4PROTO, Data: []byte{c.L4Proto}},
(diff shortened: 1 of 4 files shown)


stapelberg:
The GitHub Actions run failed:

```
2024-09-02T16:48:53.2913363Z === RUN   TestCtExpect
2024-09-02T16:48:53.3286299Z     nftables_test.go:1508: conn.Receive: netlink receive: invalid argument
2024-09-02T16:48:53.3288350Z --- FAIL: TestCtExpect (0.04s)
```

(It’s running on Ubuntu 22.04)

turekt:
Hi @stapelberg, do you maybe know which nftables version is running there? I am unable to replicate the error. I have ran tests on xubuntu 22.04.4, ubuntu 22.04 and latest debian - all are passing.

```sh
# uname -a
Linux xubuntu 6.5.0-18-generic #18~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Feb  7 18:40:03 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
# nft --version
nftables v1.0.2 (Lester Gooch)
# git branch
  master
* obj-ct-expect
# go test -run_system_tests ./...
PASS
ok      github.com/google/nftables  0.217s
---
# uname -a
Linux ubuntu 5.15.0-25-generic #25-Ubuntu SMP Wed Mar 30 15:54:22 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
# nft --version
nftables v1.0.2 (Lester Gooch)
# git branch
  master
* obj-ct-expect
# go test -run_system_tests ./...
PASS
ok      github.com/google/nftables  0.095s
---
# uname -a
Linux debian 6.1.0-23-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.99-1 (2024-07-15) x86_64 GNU/Linux
# git branch
  master
* obj-ct-expect
# nft --version
nftables v1.0.6 (Lester Gooch #5)
# go test -run_system_tests ./...
PASS
ok  	github.com/google/nftables	0.291s
```

stapelberg:
I poked around a little more in my personal repository. From the logs:

```
2024-09-05T06:51:57.5415350Z Image: ubuntu-22.04
2024-09-05T06:51:57.5415723Z Version: 20240901.1.0
2024-09-05T06:51:57.5416815Z Included Software: https://github.com/actions/runner-images/blob/ubuntu22/20240901.1/images/ubuntu/Ubuntu2204-Readme.md
2024-09-05T06:51:57.5418239Z Image Release: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240901.1
```

(I also tried on an Ubuntu 22.04 machine and couldn’t reproduce it there.)

I attached the full log, including an strace capture of the failing test: 
[job-logs.txt](https://github.com/user-attachments/files/16885209/job-logs.txt)

turekt:
This seems to be an interesting case of different handling of l3proto parameter depending on nft/kernel versions. Further analysis rendered the following results:

Test case # | Table family | L3Proto value | Unmarshal result in Github runner image | Unmarshal result in latest Debian |
| --- | --- | --- | --- | --- |
| 1 | nftables.TableFamilyINet | NFPROTO_IPV4 | FAIL: netlink receive: invalid argument | PASS |
| 2 | nftables.TableFamilyINet | _unspecified_ | PASS | FAIL: object l3proto not equal: got 1, want 0 (obviously expects NFPROTO_INET, derived from table family - let's specify NFPROTO_INET in next case to fix) |
| 3 | Inftables.TableFamilyNet | NFPROTO_INET | PASS | FAIL: netlink receive: address family not supported by protocol |
| 4 | nftables.TableFamilyIPv4 | NFPROTO_IPV4 | PASS | PASS |

Note that what manifests with `go test`, also manifests with `nft add ct expectation` command:
```
Error: Could not process rule: Invalid argument
add ct expectation inet filter expect {protocol tcp ; dport 53 ; timeout 2s ; size 1 ; l3proto ip ;}
```
The above case is when table family inet is used in combination with l3proto ipv4, confirming that the issue is not related to the nftables go lib.

As per nftables man page:
> l3proto is derived from the table family by default.

We clearly see that there is a discrepancy in how Github runner nftables handles this parameter opposed to latest OS versions. To conclude: one should always either omit the l3proto parameter or match the l3proto parameter with table family in order to get consistent results.

I have changed the initial test and specified both l3proto and table family as IPv4 as this seems to work for all versions.

stapelberg (approved):

stapelberg:
Thanks for the investigation and figuring out the issue!

PR #275, opened 2024-09-09
Added ct timeout obj

turekt:
NamedObj continuation - ct timeout stateful obj


stapelberg (approved):

stapelberg:
Thanks for the PR! CI fails, saying that the code isn’t gofmt'ed

turekt:
Should be resolved now.

stapelberg (approved):

Issue #215, opened 2023-01-15
nftables go dynset implementation will not work with libnftnl versions <1.1.9


turekt:
Hi,

I have noticed that go nftables lib tests are failing on older versions of nftables (or more precisely Debian):
```
# go test -run_system_tests ./...
--- FAIL: TestDynsetWithOneExpression (0.03s)
    nftables_test.go:3912: c.Flush() failed: conn.Receive: netlink receive: operation not supported
    nftables_test.go:3921: unexpected number of rules: got 0, want 1
--- FAIL: TestDynsetWithMultipleExpressions (0.04s)
    nftables_test.go:4018: c.Flush() failed: conn.Receive: netlink receive: operation not supported
    nftables_test.go:4027: unexpected number of rules: got 0, want 1
FAIL
exit status 1
FAIL	github.com/google/nftables	1.211s
# nft --version
nftables v0.9.8 (E.D.S.)
```

Everything seems to be fine when running tests on latest nftables/Debian: 
```
# go test -run_system_tests ./...
PASS
ok  	github.com/google/nftables	1.263s
# nft --version
nftables v1.0.2 (Lester Gooch)
```

I have traced the issue being caused by NFT_DYNSET_F_EXPR flag set here: 
https://github.com/google/nftables/blob/9aa6fdf5a28cdfec5695ebbf7d7add61eccdf61a/expr/dynset.go#L68

The libnftnl package listed on https://sources.debian.org/src/libnftnl/ show the following versions:
```
version[ 1.2.4-2](https://sources.debian.org/src/libnftnl/1.2.4-2/) (main) [bookworm, sid]
version[ 1.1.9-1](https://sources.debian.org/src/libnftnl/1.1.9-1/) (main) [bullseye]
version[ 1.1.7-1~bpo10+1](https://sources.debian.org/src/libnftnl/1.1.7-1~bpo10%2B1/) (main) [buster-backports]
version[ 1.1.2-2](https://sources.debian.org/src/libnftnl/1.1.2-2/) (main) [buster]
version[ 1.1.1-1~bpo9+1](https://sources.debian.org/src/libnftnl/1.1.1-1~bpo9%2B1/) (main) [stretch-backports]
version[ 1.0.7-1](https://sources.debian.org/src/libnftnl/1.0.7-1/) (main) [stretch]
version[ 1.0.7-1~bpo8+1](https://sources.debian.org/src/libnftnl/1.0.7-1~bpo8%2B1/) (main) [jessie-backports]
```

[...]

PR #276, opened 2024-09-11
Fix: NFT_DYNSET_F_EXPR not supported for kernels < 5.11-rc3

Ignatella:
Hi! 

We noticed that `NFT_DYNSET_F_EXPR` flag added in [v5.11-rc3](https://elixir.bootlin.com/linux/v5.11-rc3/source/include/uapi/linux/netfilter/nf_tables.h#L711)  is also used with kernels which don't support [NFTA_DYNSET_EXPRESSIONS](https://elixir.bootlin.com/linux/v5.11-rc1/source/net/netfilter/nft_dynset.c#L259) (`<v5.11-rc1`). The flag was added in the [patchset](https://lore.kernel.org/all/20210103192920.18639-1-pablo@netfilter.org/#R) not to fail silently. 

The described [behavior](https://elixir.bootlin.com/linux/v5.10.225/source/net/netfilter/nft_dynset.c#L129) causes error for older kernels, what this contribution aims to fix.

### DIFF
--- a/expr/dynset.go
+++ b/expr/dynset.go
@@ -77,7 +77,6 @@ func (e *Dynset) marshalData(fam byte) ([]byte, error) {
 
 	// Per https://git.netfilter.org/libnftnl/tree/src/expr/dynset.c?id=84d12cfacf8ddd857a09435f3d982ab6250d250c#n170
 	if len(e.Exprs) > 0 {
-		flags |= NFT_DYNSET_F_EXPR
 		switch len(e.Exprs) {
 		case 1:
 			exprData, err := Marshal(fam, e.Exprs[0])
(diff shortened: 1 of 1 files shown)


stapelberg:
@turekt As the original contributor of the dynset code, can you take a look at this PR please? Thank you.

turekt:
Hi @stapelberg and @Ignatella,

this is a known issue previously discussed in #215. I have checked the original code and it seems that NFT_DYNSET_F_EXPR flag is set only when multiple expressions are used in dynset (see https://git.netfilter.org/nftables/tree/src/netlink_linearize.c?id=036a1eb3304995e4e000c552b9dfd33b1073ddb6#n1566) so this PR makes sense.

Note that this will fix support for single expressions on older kernels but multiple expressions on older kernels will remain unsupported as NFT_DYNSET_F_EXPR flag should not be omitted for dynsets with multiple expressions.

LGTM

PR #277, opened 2024-09-13
Add: netlink socket options

Ignatella:
Hi!

While using nftables we understood that it would be great to be able to set some options on the socket, i.e. setting r/w buffer size or `pid` retrieval to track what process issues rule changes. 

This PR attemps to add such api to the `New` method through options (callbacks) executed every time `dialNetlink` called.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -37,15 +37,19 @@ type Conn struct {
 	TestDial nltest.Func // for testing only; passed to nltest.Dial
 	NetNS    int         // fd referencing the network namespace netlink will interact with.
 
-	lasting  bool       // establish a lasting connection to be used across multiple netlink operations.
-	mu       sync.Mutex // protects the following state
-	messages []netlink.Message
-	err      error
-	nlconn   *netlink.Conn // netlink socket using NETLINK_NETFILTER protocol.
+	lasting     bool       // establish a lasting connection to be used across multiple netlink operations.
+	mu          sync.Mutex // protects the following state
+	messages    []netlink.Message
+	err         error
+	nlconn      *netlink.Conn // netlink socket using NETLINK_NETFILTER protocol.
+	sockOptions []SockOption
 }
 
 // ConnOption is an option to change the behavior of the nftables Conn returned by Open.
 type ConnOption func(*Conn)
+
+// SockOption is an option to change the behavior of the netlink socket used by the nftables Conn.
+type SockOption func(*netlink.Conn) error
 
 // New returns a netlink connection for querying and modifying nftables. Some
 // aspects of the new netlink connection can be configured using the options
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -284,11 +296,30 @@ func (cc *Conn) FlushRuleset() {
 }
 
 func (cc *Conn) dialNetlink() (*netlink.Conn, error) {
+	var (
+		conn *netlink.Conn
+		err  error = nil
```
stapelberg, line 301: There is no need to explicitly initialize variables in Go. 

Variables will always be initialized to their zero value by default.

This can just be `err error`, without the `= nil`

conn.go
```diff
@@ -284,11 +296,30 @@ func (cc *Conn) FlushRuleset() {
...
+	} else {
+		conn, err = netlink.Dial(unix.NETLINK_NETFILTER, &netlink.Config{NetNS: cc.NetNS})
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	for _, opt := range cc.sockOptions {
+		err := opt(conn)
```
stapelberg, line 315: Merge this function call into the if conditional:

```go
if err := opt(conn); err != nil {
  return nil, err
}
```

stapelberg (changes requested):

stapelberg:
> While using nftables we understood that it would be great to be able to set some options on the socket, i.e. setting r/w buffer size or `pid` retrieval to track what process issues rule changes.

Out of curiosity: Under which conditions do you need to adjust the r/w buffer size? 

Is the code which sets the options open source? If so, can you share a link please?

Ignatella:
@stapelberg 

> Out of curiosity: Under which conditions do you need to adjust the r/w buffer size?

One of the flows and assumptions we must consider is the ability to set up to 5,000 rules in a single transaction. However, using the default buffer sizes causes the process to get stuck indefinitely.

```go
conn, err: = nftables.New(nftables.WithSockOptions(func(conn * netlink.Conn) error {
    // 20 mb for read and write buffer
    if err: = conn.SetReadBuffer(20971520);
    err != nil {
        return err
    }

    if err: = conn.SetWriteBuffer(20971520);
    err != nil {
        return err
    }

    pid, err: = getPortID(conn)

    if err != nil {
        return err
    }

    nftportid.SetNftablesPortID(pid)

    return nil
}))


func getPortID(conn * netlink.Conn)(uint32, error) {
    var (
        sa unix.Sockaddr err error
    )
    rawConn, err: = conn.SyscallConn()
    if err != nil {
        return 0, err
    }
    doErr: = rawConn.Control(func(fd uintptr) {
        for {
            sa, err = unix.Getsockname(int(fd))
            if ready(err) {
                return
            }
        }
    })
    if doErr != nil {
        return 0, doErr
    }
    if err != nil {
        return 0, err
    }
    saCast, ok: = sa.( * unix.SockaddrNetlink)
    if !ok {
        return 0, errors.New("SockaddrNetlink missing? should not happen")
    }
    return saCast.Pid, nil
}
```
The code is not open source, but we can share how we use the proposed functionality. 😊 Also, please note the 'port ID' part, which we use to track who issues the changes.

stapelberg (approved):

PR #283, opened 2024-10-16
Update: process monitor events all at once upon commit

Ignatella:
Hi!

While using nftables, we discovered it would be more efficient to process monitor events (changes) all at once upon commit, triggered by the event that caused the change, such as [NFT_MSG_NEWGEN](https://elixir.bootlin.com/linux/v6.10.10/source/net/netfilter/nf_tables_api.c#L10514).

This PR modifies the way events are handled by grouping all changes together and processing them in response to the event that initiated the change. Additionally, the event header, which includes important information such as the PID, is now propagated. This allows us to identify the process responsible for the change.

### DIFF
--- a/monitor.go
+++ b/monitor.go
@@ -302,7 +302,7 @@ func (monitor *Monitor) monitor() {
 					Changes:     changesEvents,
 				}
 
-				changesEvents = make([]*MonitorEvent, 0)
+				changesEvents = nil
 			}
 		}
 	}
--- a/monitor.go
+++ b/monitor.go
@@ -198,7 +198,7 @@ func NewMonitor(opts ...MonitorOption) *Monitor {
 }
 
 func (monitor *Monitor) monitor() {
-	changesEvents := make([]*MonitorEvent, 0)
+	var changesEvents []*MonitorEvent
 
 	for {
 		msgs, err := monitor.conn.Receive()
@@ -220,7 +220,7 @@ func (monitor *Monitor) monitor() {
 					GeneratedBy: event,
 					Changes:     changesEvents,
 				}
-				changesEvents = make([]*MonitorEvent, 0)
+				changesEvents = nil
 
 				break
 			}
@@ -333,7 +333,6 @@ func (monitor *Monitor) Close() error {
 // handle, for now.
 func (cc *Conn) AddMonitor(monitor *Monitor) (chan *MonitorEvent, error) {
 	generationalEventCh, err := cc.AddGenerationalMonitor(monitor)
-
 	if err != nil {
 		return nil, err
 	}
--- a/monitor.go
+++ b/monitor.go
@@ -116,9 +116,15 @@ const (
 // nftables.MonitorEventTypeNewTable, you can access the corresponding table
 // details via Data.(*nftables.Table).
 type MonitorEvent struct {
-	Type  MonitorEventType
-	Data  any
-	Error error
+	Header netlink.Header
+	Type   MonitorEventType
+	Data   any
+	Error  error
+}
+
+type MonitorEvents struct {
+	GenerateBy *MonitorEvent
+	Changes    []*MonitorEvent
 }
 
 const (
@@ -139,7 +145,7 @@ type Monitor struct {
 
 	// mu covers eventCh and status
 	mu      sync.Mutex
-	eventCh chan *MonitorEvent
+	eventCh chan *MonitorEvents
 	status  int
 }
 
@@ -147,7 +153,7 @@ type MonitorOption func(*Monitor)
 
 func WithMonitorEventBuffer(size int) MonitorOption {
 	return func(monitor *Monitor) {
-		monitor.eventCh = make(chan *MonitorEvent, size)
+		monitor.eventCh = make(chan *MonitorEvents, size)
 	}
 }
 
@@ -177,7 +183,7 @@ func NewMonitor(opts ...MonitorOption) *Monitor {
 		opt(monitor)
 	}
 	if monitor.eventCh == nil {
-		monitor.eventCh = make(chan *MonitorEvent)
+		monitor.eventCh = make(chan *MonitorEvents)
 	}
 	objects, ok := monitorFlags[monitor.action]
 	if !ok {
@@ -192,6 +198,8 @@ func NewMonitor(opts ...MonitorOption) *Monitor {
 }
 
 func (monitor *Monitor) monitor() {
+	changesEvents := make([]*MonitorEvent, 0, 2)
+
 	for {
 		msgs, err := monitor.conn.Receive()
 		if err != nil {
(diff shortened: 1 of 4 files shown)


set.go
```diff
@@ -797,10 +803,16 @@ func elementsFromMsg(fam byte, msg netlink.Message) ([]SetElement, error) {
 	}
 	ad.ByteOrder = binary.BigEndian
 
+	var info = &SetElementsInfo{}
```
patryk4815, line 806: missing return?

monitor.go
```diff
@@ -293,7 +331,7 @@ func (monitor *Monitor) Close() error {
 // calling Close on Monitor or encountering a netlink conn error while Receive.
 // Caller may receive a MonitorEventTypeOOB event which contains an error we didn't
 // handle, for now.
-func (cc *Conn) AddMonitor(monitor *Monitor) (chan *MonitorEvent, error) {
+func (cc *Conn) AddMonitor(monitor *Monitor) (chan *MonitorEvents, error) {
```
stapelberg, line 334: There are existing users of this method, so we should not change its API.

Maybe we could introduce a new AddGenerationalMonitor() method with your suggested signature, and make AddMonitor work with generations, unpack the events, put them on a channel of the old type?

monitor.go
```diff
@@ -192,20 +198,30 @@ func NewMonitor(opts ...MonitorOption) *Monitor {
 }
 
 func (monitor *Monitor) monitor() {
+	changesEvents := make([]*MonitorEvent, 0, 2)
```
stapelberg, line 201: Here and below: Why initialize the slice with capacity 2? It seems unlikely to me that this would make a difference. The slice’s capacity will quickly grow to the maximum needed and then stay there, I think.

monitor.go
```diff
@@ -116,9 +116,15 @@ const (
...
-	Data  any
-	Error error
+	Header netlink.Header
+	Type   MonitorEventType
+	Data   any
+	Error  error
+}
+
+type MonitorEvents struct {
+	GenerateBy *MonitorEvent
```
stapelberg, line 126: I think `GeneratedBy` (with a d) reads better.

set.go
```diff
@@ -268,6 +268,12 @@ type Set struct {
 	KeyByteOrder binaryutil.ByteOrder
 }
 
+type SetElementsInfo struct {
```
stapelberg, line 271: What’s the relation between this part of the change and the rest of it? Could the set changes go into a separate commit / PR?

stapelberg (changes requested):

stapelberg:
cc @singchia who contributed the monitoring code

set.go
Ignatella: yes, thank You

monitor.go
Ignatella: AddGenerationalMonitor introduced, thank You

monitor.go
Ignatella: switched to

```go
make([]*MonitorEvent, 0)
```

monitor.go
Ignatella: switched to `GeneratedBy`

set.go
Ignatella: Removed from the PR, I will propose this change as part of another PR. Thank You

monitor.go
```diff
@@ -192,20 +198,30 @@ func NewMonitor(opts ...MonitorOption) *Monitor {
 }
 
 func (monitor *Monitor) monitor() {
+	changesEvents := make([]*MonitorEvent, 0)
```
stapelberg, line 201: Write `var changesEvents []*MonitorEvent` here. `make(…, 0)` is equivalent to just declaring a variable.

monitor.go
```diff
@@ -192,20 +198,30 @@ func NewMonitor(opts ...MonitorOption) *Monitor {
...
 				}
-				monitor.eventCh <- event
+
+				changesEvents = append(changesEvents, event)
+
+				monitor.eventCh <- &MonitorEvents{
+					GeneratedBy: event,
+					Changes:     changesEvents,
+				}
+				changesEvents = make([]*MonitorEvent, 0)
```
stapelberg, line 223: This can just be `changesEvents = nil`

monitor.go
```diff
@@ -294,6 +332,27 @@ func (monitor *Monitor) Close() error {
 // Caller may receive a MonitorEventTypeOOB event which contains an error we didn't
 // handle, for now.
 func (cc *Conn) AddMonitor(monitor *Monitor) (chan *MonitorEvent, error) {
+	generationalEventCh, err := cc.AddGenerationalMonitor(monitor)
+
```
stapelberg, line 336: nit: remove blank line between err assignment and err check

stapelberg (changes requested):

monitor.go
Ignatella: changed to declaration. Thank you

monitor.go
Ignatella: Changed to `changesEvents = nil`

monitor.go
Ignatella: Removed

Ignatella:
Also added ProcComm check to `TestMonitor`

monitor_test.go
```diff
@@ -66,23 +69,35 @@ func TestMonitor(t *testing.T) {
 			if !ok {
 				return
 			}
-			if event.Error != nil {
-				err = fmt.Errorf("monitor err: %s", event.Error)
```
patryk4815, line 70: the original version of the test seems to be broken, "err" is not checked at all

monitor_test.go
```diff
@@ -66,23 +69,35 @@ func TestMonitor(t *testing.T) {
...
+					gotTable = change.Data.(*nftables.Table)
+					atomic.AddInt32(&count, 1)
+				case nftables.MonitorEventTypeNewChain:
+					gotChain = change.Data.(*nftables.Chain)
+					atomic.AddInt32(&count, 1)
+				case nftables.MonitorEventTypeNewRule:
+					gotRule = change.Data.(*nftables.Rule)
+					atomic.AddInt32(&count, 1)
+				}
+				if atomic.LoadInt32(&count) == 3 {
```
patryk4815, line 98: this check looks bad to me, should throw error when count != 3

monitor.go
```diff
@@ -221,54 +237,76 @@ func (monitor *Monitor) monitor() {
...
+					Header: msg.Header,
 				}
-				monitor.eventCh <- event
+
+				monitor.eventCh <- &MonitorEvents{
+					GeneratedBy: event,
+					Changes:     changesEvents,
+				}
+
+				changesEvents = make([]*MonitorEvent, 0)
```
stapelberg, line 305: This should also be `changesEvents = nil`

stapelberg (commented):
Almost ready to merge

monitor.go
Ignatella: Changed, thank You

stapelberg (approved):

PR #287, opened 2024-12-10
Fix incorrect size check in NFGenMsg

patryk4815:
This patch resolves an issue with the size check in `NFGenMsg`.

Currently, the size is truncated to 4 bytes, but the check incorrectly expects 16 bytes. As a result, `NFGenMsg.Decode` is never executed.

You can find the relevant code here:
https://github.com/google/nftables/blob/51c44dcf0573a890bb26fc621fa0142e16b94261/monitor.go#L375

This fix ensures that NFGenMsg is decoded correctly.


patryk4815:
@stapelberg done ;)

stapelberg (approved):

PR #292, opened 2025-01-04
Add integration tests for nftables package

aojea:
Adds integration tests to the nftables package to verify
that the Go code correctly programs nftables rules. The tests use
external nftables scripts to define the expected state and compare
it with the state produced by the Go code.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -33,3 +33,5 @@ jobs:
         go test ./...
         go test -c github.com/google/nftables
         sudo ./nftables.test -test.v -run_system_tests
+        go test -c github.com/google/nftables/integration
+        (cd integration && sudo ../integration.test -test.v -run_system_tests)
--- a/go.mod
+++ b/go.mod
@@ -3,13 +3,14 @@ module github.com/google/nftables
 go 1.21
 
 require (
+	github.com/google/go-cmp v0.6.0
 	github.com/mdlayher/netlink v1.7.3-0.20250113171957-fbb4dce95f42
-	github.com/vishvananda/netns v0.0.0-20180720170159-13995c7128cc
+	github.com/vishvananda/netlink v1.3.0
+	github.com/vishvananda/netns v0.0.4
 	golang.org/x/sys v0.28.0
 )
 
 require (
-	github.com/google/go-cmp v0.6.0 // indirect
 	github.com/mdlayher/socket v0.5.0 // indirect
 	golang.org/x/net v0.33.0 // indirect
 	golang.org/x/sync v0.6.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -4,11 +4,15 @@ github.com/mdlayher/netlink v1.7.3-0.20250113171957-fbb4dce95f42 h1:A1Cq6Ysb0GM0
 github.com/mdlayher/netlink v1.7.3-0.20250113171957-fbb4dce95f42/go.mod h1:BB4YCPDOzfy7FniQ/lxuYQ3dgmM2cZumHbK8RpTjN2o=
 github.com/mdlayher/socket v0.5.0 h1:ilICZmJcQz70vrWVes1MFera4jGiWNocSkykwwoy3XI=
 github.com/mdlayher/socket v0.5.0/go.mod h1:WkcBFfvyG8QENs5+hfQPl1X6Jpd2yeLIYgrGFmJiJxI=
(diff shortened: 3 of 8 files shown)


tests/nft_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+			// Real interface must exist otherwise some nftables will fail
+			la := netlink.NewLinkAttrs()
+			la.Name = "dummy0"
+			dummy := &netlink.Dummy{LinkAttrs: la}
+			if err := netlink.LinkAdd(dummy); err != nil {
+				t.Fatal(err)
+			}
+
+			_, filename, _, _ := runtime.Caller(0) // Get the current file's path
```
stapelberg, line 193: This hack should be entirely unnecessary. Go tests are run with their working directory set such that testdata/… paths work.

tests/nft_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+			// Retrieve nftables state using nft
+			expectedOutput, err := getNFTablesRuleset()
+			if err != nil {
+				t.Fatalf("Failed to list nftables ruleset: %v\noutput:%s", err, expectedOutput)
+			}
+			t.Logf("Expected output:\n%s", expectedOutput)
+
+			// Program nftables using your Go code
+			err = flushNFTRuleset()
```
stapelberg, line 213: here and below: nit: if err := flushNFTRuleset(); err != nil {

tests/nft_test.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	return normalizeOutput(string(out)), nil
+}
+
+func flushNFTRuleset() error {
+	cmd := exec.Command("nft", "flush", "ruleset")
+	return cmd.Run()
+}
+
+func normalizeOutput(output string) string {
```
stapelberg, line 262: Instead of defining this function, make the callers call strings.TrimSpace() directly

tests/nft_test.go
```diff
@@ -0,0 +1,265 @@
...
+func applyNFTScript(scriptPath string) (string, error) {
+	cmd := exec.Command("nft", "-f", scriptPath)
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return string(out), err
+	}
+	return normalizeOutput(string(out)), nil
+}
+
+func getNFTablesRuleset() (string, error) {
```
stapelberg, line 248: nit: rename to listNFTRuleset

tests/nft_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+			err = flushNFTRuleset()
+			if err != nil {
+				t.Fatalf("Failed to flush nftables ruleset: %v", err)
+			}
+		})
+	}
+}
+
+func applyNFTScript(scriptPath string) (string, error) {
```
stapelberg, line 239: nit: rename to applyNFTRuleset (the nft(8) manpage never mentions “script” and only talks about rulesets)

.github/workflows/push.yml
```diff
@@ -33,3 +33,5 @@ jobs:
         go test ./...
         go test -c github.com/google/nftables
         sudo ./nftables.test -test.v -run_system_tests
+        go test -c github.com/google/nftables/tests
```
stapelberg, line 36: Can you rename the directory from tests to integration please?

stapelberg (changes requested):
Thanks! I originally thought “wait, we already have such an integration test!”, but then I realized it’s not in this repository: The test I had in mind was https://github.com/rtr7/router7/blob/master/integration/netconfig/netconfig_test.go from https://router7.org/

Just a few small changes, very nice overall!

aojea:
addressed comments

stapelberg (approved):

.github/workflows/push.yml
```diff
@@ -33,3 +33,5 @@ jobs:
         go test ./...
         go test -c github.com/google/nftables
         sudo ./nftables.test -test.v -run_system_tests
+        go test -c github.com/google/nftables/integration
+        sudo ./integration.test -test.v -run_system_tests
```
stapelberg, line 37: The tests succeed when run with the `go` tool, but not on GitHub Actions.

I think the best fix is to change directory in this line:

```
(cd integration && sudo ../integration.test -test.v -run_system_tests)
```

aojea: done

stapelberg (approved):

PR #290, opened 2024-12-18
Add support for set comments

turekt:
Per issue #288, I've added a simple change to sets to support adding comments.

PR #294, opened 2025-01-11
set: Add set support for size specifier, add missing dynamic flag unmarshal

a-ilin:
Add support for set size specification: handle attribute `NFTNL_SET_DESC_SIZE`, as done in `libnftnl`: https://git.netfilter.org/libnftnl/tree/src/set.c#n424

Example:
```sh
nft add set ip filter myset { type ipv4_addr\; size 65535\; flags dynamic\; }
```

Also add missing unmarshal of `Dynamic` flag (`NFT_SET_EVAL`).

### DIFF
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -4103,6 +4103,46 @@ func TestSetElementsInterval(t *testing.T) {
 	}
 }
 
+func TestSetSizeConcat(t *testing.T) {
+	// Create a new network namespace to test these operations,
+	// and tear down the namespace at test completion.
+	c, newNS := nftest.OpenSystemConn(t, *enableSysTests)
+	defer nftest.CleanupSystemConn(t, newNS)
+	// Clear all rules at the beginning + end of the test.
+	c.FlushRuleset()
+	defer c.FlushRuleset()
+
+	filter := c.AddTable(&nftables.Table{
+		Family: nftables.TableFamilyIPv6,
+		Name:   "filter",
+	})
+
+	set := &nftables.Set{
+		Name:          "test-set",
+		Table:         filter,
+		KeyType:       nftables.MustConcatSetType(nftables.TypeIP6Addr, nftables.TypeInetService, nftables.TypeIP6Addr),
+		Dynamic:       true,
+		Concatenation: true,
+		Size:          200,
+	}
+
+	if err := c.AddSet(set, nil); err != nil {
+		t.Errorf("c.AddSet(set) failed: %v", err)
+	}
+
+	if err := c.Flush(); err != nil {
+		t.Errorf("c.Flush() failed: %v", err)
+	}
+
+	sets, err := c.GetSets(filter)
+	if err != nil {
+		t.Errorf("c.GetSets() failed: %v", err)
+	}
+	if len(sets) != 1 {
+		t.Fatalf("len(sets) = %d, want 1", len(sets))
+	}
+}
+
 func TestCreateListFlowtable(t *testing.T) {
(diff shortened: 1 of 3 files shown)


stapelberg:
@turekt You contributed support for set comments in PR https://github.com/google/nftables/pull/290 — could you take a look at this PR and confirm everything looks good please?

turekt (commented):
I would personally add an additional system test which checks if Size works ok in correlation with NFTA_SET_DESC_CONCAT. We work with this field when set uses a concatenation and this seems to be a part of the set element description https://git.netfilter.org/libnftnl/tree/include/linux/netfilter/nf_tables.h?id=be0bae0ad31b0adb506f96de083f52a2bd0d4fbf#n326. 

I did a quick test on sample code and it seems to work ok. Here's sample code I've used
```go
conn, err := nftables.New()
if err != nil {
	panic(err)
}
conn.FlushRuleset()
table := &nftables.Table{
	Family: nftables.TableFamilyIPv4,
	Name:   "filter",
}
conn.AddTable(table)

set := &nftables.Set{
	Name:    "test-set",
	Table:   table,
	KeyType: nftables.MustConcatSetType(nftables.TypeIP6Addr, nftables.TypeInetService, nftables.TypeIP6Addr),
	Dynamic: true,
	Concatenation: true,
	Size: 200,
}
if err := conn.AddSet(set, nil); err != nil {
	panic(err)
}
if err := conn.Flush(); err != nil {
	panic(err)
}
```

There is also an example test here https://github.com/google/nftables/blob/85aee131ff7020abd2d5abb7268f00c517152757/nftables_test.go#L4036

Otherwise implementation looks ok.

a-ilin:
Hi @turekt , 

Thanks for the review!

Please find the mentioned test added in the new commit: https://github.com/google/nftables/pull/294/commits/a6f09ce2546582b5aca979b3a5c908922bcae829

nftables_test.go
```diff
@@ -4103,6 +4103,46 @@ func TestSetElementsInterval(t *testing.T) {
...
+		Name:   "filter",
+	})
+
+	set := &nftables.Set{
+		Name:    "test-set",
+		Table:   filter,
+		KeyType: nftables.MustConcatSetType(nftables.TypeIP6Addr, nftables.TypeInetService, nftables.TypeIP6Addr),
+		Dynamic: true,
+		Concatenation: true,
+		Size: 200,
```
turekt, line 4126: Misalignment suggests that code is not formatted as per gofmt. This should be done since Github action checks this.
https://github.com/google/nftables/blob/1c789726cfebc8afac9383f7005f381f05959b37/.github/workflows/push.yml#L26-L28

a-ilin: You are right, sorry! I re-pushed the last commit.

turekt (approved):

set_test.go
```diff
@@ -185,3 +189,82 @@ func TestConcatSetTypeElements(t *testing.T) {
...
+			name: "Set with size, timeout, dynamic flag specified",
+			set: Set{
+				Name:       "test-set",
+				ID:         uint32(2),
+				HasTimeout: true,
+				Dynamic:    true,
+				Size:       10,
+				Table:      tbl,
+				KeyType:    TypeIPAddr,
+				Timeout:    time.Duration(30) * time.Second,
```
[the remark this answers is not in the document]
a-ilin: Fixed, with the commits related to `Size` property squashed. Please check: https://github.com/google/nftables/compare/28fb497ab8341973a1a19048089d983e5123f1f8..fcdf440964465db87aaceee41fd3fff74e44eb5e

stapelberg (approved):

Issue #291, opened 2025-01-02
FIB objects not present when reading


aojea:
Repro

```
nft --debug=netlink insert rule inet cni-kindnet output fib daddr type local
inet cni-kindnet output
  [ fib daddr type => reg 1 ]
  [ cmp eq reg 1 0x00000002 ]
```

gives
```
nft list chain inet cni-kindnet output
table inet cni-kindnet {
        chain output {
                type nat hook output priority -100; policy accept;
                fib daddr type local
        }
}
```

Using my "hacky" golang dumper https://gist.github.com/aojea/8a811e11a060fa638aa2dd8165960d5a
```
/nfroundtrip cni-kindnet output
2025/01/02 15:51:48 rule position 0
&expr.Cmp{Op:0x0, Register:0x1, Data:[]uint8{0x2, 0x0, 0x0, 0x0}}
```

right expressions with golang

```
		// only packets destined to local addresses
		&expr.Fib{
			Register:       1,
			FlagDADDR:      true,
			ResultADDRTYPE: true,
		},
		&expr.Cmp{
			// [ cmp eq reg 1 0x00000002 ]
			Op:       expr.CmpOpEq,
			Register: 1,
			Data:     encodeWithAlignment(byte(unix.RTN_LOCAL)),
		},
```

so it seems the parsing is not working as expected? it is missing the FIB object

turekt:
Good catch. I fixed it by adding expr name to `exprFromName` function.
```
# go run main.go cni-kindnet output
2025/01/15 00:37:58 rule position 0
&expr.Fib{Register:0x1, ResultOIF:true, ResultOIFNAME:false, ResultADDRTYPE:false, FlagSADDR:false, FlagDADDR:false, FlagMARK:false, FlagIIF:false, FlagOIF:false, FlagPRESENT:false}
&expr.Cmp{Op:0x0, Register:0x1, Data:[]uint8{0x2, 0x0, 0x0, 0x0}}
```

As you can see, this still did not produce the correct result so I made a few changes to fib marshaling as well. Now it seems to work as expected:
```
# go run dumper.go filter test-chain
2025/01/15 01:38:45 rule position 0
&expr.Fib{Register:0x1, ResultOIF:false, ResultOIFNAME:false, ResultADDRTYPE:true, FlagSADDR:false, FlagDADDR:true, FlagMARK:false, FlagIIF:false, FlagOIF:false, FlagPRESENT:false}
[...]

PR #296, opened 2025-01-14
Fix Fib parsing

turekt:
Per issue #291, FIB objects are not parsed although the implementation exists. This PR fixes this issue.

Additionally, the initial FIB implementation did not parse the object properly so I changed that as well. The initial implementation assumed that result field is uint32 used as flag, but per original libnftnl implementation, I would say that result is uint32 referring to only a single value: https://git.netfilter.org/libnftnl/tree/include/linux/netfilter/nf_tables.h?id=ff37c01480cd0b938658d180ff5c7b1958ad250f#n1605

Since the initial implementation used booleans, I decided to not change that for this PR.

### DIFF
--- a/expr/fib.go
+++ b/expr/fib.go
@@ -118,17 +118,22 @@ func (e *Fib) unmarshal(fam byte, data []byte) error {
 			e.Register = ad.Uint32()
 		case unix.NFTA_FIB_RESULT:
 			result := ad.Uint32()
-			e.ResultOIF = (result & unix.NFT_FIB_RESULT_OIF) == 1
-			e.ResultOIFNAME = (result & unix.NFT_FIB_RESULT_OIFNAME) == 1
-			e.ResultADDRTYPE = (result & unix.NFT_FIB_RESULT_ADDRTYPE) == 1
+			switch result {
+			case unix.NFT_FIB_RESULT_OIF:
+				e.ResultOIF = true
+			case unix.NFT_FIB_RESULT_OIFNAME:
+				e.ResultOIFNAME = true
+			case unix.NFT_FIB_RESULT_ADDRTYPE:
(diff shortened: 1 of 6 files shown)


expr/fib.go
```diff
@@ -118,17 +118,22 @@ func (e *Fib) unmarshal(fam byte, data []byte) error {
...
+			}
 		case unix.NFTA_FIB_FLAGS:
 			flags := ad.Uint32()
-			e.FlagSADDR = (flags & unix.NFTA_FIB_F_SADDR) == 1
-			e.FlagDADDR = (flags & unix.NFTA_FIB_F_DADDR) == 1
-			e.FlagMARK = (flags & unix.NFTA_FIB_F_MARK) == 1
-			e.FlagIIF = (flags & unix.NFTA_FIB_F_IIF) == 1
-			e.FlagOIF = (flags & unix.NFTA_FIB_F_OIF) == 1
-			e.FlagPRESENT = (flags & unix.NFTA_FIB_F_PRESENT) == 1
+			e.FlagSADDR = (flags & unix.NFTA_FIB_F_SADDR) == unix.NFTA_FIB_F_SADDR
```
stapelberg, line 131: Another way of expressing this check is `(flags & unix.NFTA_FIB_F_SADDR) != 0`, right?

Could you change this code to the `!= 0` route for consistency with set.go?

Are the other occurrences of '… & … == 1` correct or should they also be changed? (e.g. expr/limit.go, quota.go, expr/quota.go, AFAICT)

stapelberg (commented):
Thanks for sending this fix!

expr/fib.go
turekt: All good points. Changed as requested.

PR #299, opened 2025-02-20
Set rule handle during flush

jscissr:
This change makes it possible to delete rules after inserting them, without needing to query the rules first. Rules can be deleted both before and after they are flushed. Additionally, this allows positioning a new rule next to an existing rule, both before and after the existing rule is flushed.

This is a new and improved implementation of the feature proposed in https://github.com/google/nftables/pull/88. I tried to address the unresolved comments there. See commit messages for more details.

Closes #88

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -41,12 +41,18 @@ type Conn struct {
 
 	lasting      bool       // establish a lasting connection to be used across multiple netlink operations.
 	mu           sync.Mutex // protects the following state
-	messages     []netlink.Message
+	messages     []netlinkMessage
 	err          error
 	nlconn       *netlink.Conn // netlink socket using NETLINK_NETFILTER protocol.
 	sockOptions  []SockOption
 	lastID       uint32
 	allocatedIDs uint32
+}
+
+type netlinkMessage struct {
+	Header netlink.Header
+	Data   []byte
+	rule   *Rule
 }
 
 // ConnOption is an option to change the behavior of the nftables Conn returned by Open.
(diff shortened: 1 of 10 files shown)


stapelberg:
> This PR might be too large to review, so I will split it up into multiple smaller PRs.

Thanks, smaller PRs are better for the limited time budget I have right now for this project :)

jscissr:
Thanks for your reviews so far! This PR now contains only the remaining changes and is ready for review.

conn.go
```diff
@@ -263,15 +250,53 @@ func (cc *Conn) Flush() error {
...
+		}
+		replies = replies[1:]
+		if len(replies) == 0 {
+			replies, err = conn.Receive()
+		}
+	}
+
 	// Fetch the requested acknowledgement for each message we sent.
-	for _, msg := range cc.messages {
-		if _, err := receiveAckAware(conn, msg.Header.Flags); err != nil {
```
stapelberg, line 273: What about the other callers of `receiveAckAware`? Would they need a similar fix?

rule.go
```diff
@@ -164,20 +167,23 @@ func (cc *Conn) newRule(r *Rule, op ruleOperation) *Rule {
...
 	msgData = append(msgData, data...)
-	var flags netlink.HeaderFlags
 	if r.UserData != nil {
 		msgData = append(msgData, cc.marshalAttr([]netlink.Attribute{
 			{Type: unix.NFTA_RULE_USERDATA, Data: r.UserData},
 		})...)
 	}
 
+	var flags netlink.HeaderFlags
+	var handleReply func(reply netlink.Message) error
```
stapelberg, line 177: Is it really required to make the reply handling a function pointer? Couldn’t we inspect the reply message flags to see if this was an echo for a create operation?

conn.go
jscissr: I'm not sure which fix you mean. The other callers of `receiveAckAware` all send a single message, and then receive a single reply and possibly an ack.

Note that originally, receiveAckAware was not used here. Part of this PR is reverting commit 0d9bfa4d18da72597950d8a903b56fb1d6e9c830, which is very misleading, and doing it properly instead. As already described in the commit message of this PR, there is no such thing as an "overrun flag"; the problem which that commit was attempting to fix was that Flush was not receiving replies.

rule.go
jscissr: I used a callback because that seemed like the simplest way to do it. Would you prefer the `handleReply` field of `netlinkMessage` to have an interface type instead of a function type?

jscissr: Changed it to what you suggested.

conn.go
stapelberg: Thanks, that one is now merged. Can you rebase please?

PR #301, opened 2025-02-21
fix unmarshalling of expr.Ct source register

### LINKED ISSUE #300: expr.Ct register is not unmarshalled
With this nftables configuration:
```
table ip route-udp {
	chain raw {
		type filter hook prerouting priority raw; policy accept;
		udp dport 1000 ct zone set 1
	}
}
```

Unmarshaling should yield the following connection tracking rule:
```go
&expr.Ct{Register:0x1, SourceRegister:true, Key:0x11, Direction:0x0}
```

But it yields:
```go
&expr.Ct{Register:0x0, SourceRegister:false, Key:0x11, Direction:0x0}
```

marten-seemann:
Fixes #300.

### DIFF
--- a/expr/ct.go
+++ b/expr/ct.go
@@ -186,6 +186,9 @@ func (e *Ct) unmarshal(fam byte, data []byte) error {
 			e.Register = ad.Uint32()
 		case unix.NFTA_CT_DIRECTION:
 			e.Direction = ad.Uint32()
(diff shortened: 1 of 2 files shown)


stapelberg (approved):

stapelberg:
Thanks!

PR #303, opened 2025-02-26
Split set elements into batches if needed

jscissr:
If the number of elements to be added to or removed from a set is large, they may not all fit into one message, because the size field of a netlink attribute is a uint16 and would overflow. To support this case, the elements need to be split into multiple batches.

### DIFF
--- a/set.go
+++ b/set.go
@@ -18,6 +18,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"math"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


set.go
```diff
@@ -379,24 +379,31 @@ func (cc *Conn) SetAddElements(s *Set, vals []SetElement) error {
...
-
-	return nil
+	return cc.appendElemList(s, vals, unix.NFT_MSG_DELSETELEM)
 }
 
-func (s *Set) makeElemList(vals []SetElement, id uint32) ([]netlink.Attribute, error) {
+// maxElemBatchSize is the maximum size in bytes of encoded set elements which
+// are sent in one netlink message. The size field of a netlink attribute is a
+// uint16, and 1024 bytes is more than enough for the per-message headers.
+const maxElemBatchSize = 0x10000 - 1024
```
stapelberg, line 398: Can you explain where the 0x10000 comes from please?

stapelberg (commented):
Looks good overall

set.go
stapelberg: OK. Can we express this value in terms of `math.MaxUint16` for clarity? https://pkg.go.dev/math#pkg-constants

stapelberg (approved):

stapelberg:
Thanks!

PR #304, opened 2025-03-03
Deprecate Rule.Flags field

jscissr:
The functionality added in a46119e5 never worked: If you set NFTA_RULE_POSITION to 0, the kernel will just complain that a rule with this handle does not exist. This removes the broken functionality, leaving the field deprecated.

The right way to insert a rule at the beginning of a chain is to use InsertRule and leave Position unset.

https://github.com/google/nftables/issues/126 mentions that the nft command allows referring to rules by index. But here is a quote from the nft manpage:

> The add and insert commands support an optional location specifier,
> which is either a handle or the index (starting at zero) of an
> existing rule. Internally, rule locations are always identified by
> handle and the translation from index happens in userspace.

In other words, identifiying rules by index is a feature of nft and is not part of the kernel interface.

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -62,10 +62,8 @@ type Rule struct {
 	// PositionID can be set to the ID of another Rule, same as Position, for when
 	// the existing rule is not yet committed.
 	PositionID uint32
-	// The list of possible flags are specified by nftnl_rule_attr, see
-	// https://git.netfilter.org/libnftnl/tree/include/libnftnl/rule.h#n21
-	// Current nftables go implementation supports only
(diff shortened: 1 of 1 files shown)


stapelberg:
cc @turekt

stapelberg (approved):

stapelberg:
Can you rebase this PR please?

jscissr:
I will rebase this once #299 is closed, because that might cause another conflict otherwise.

jscissr:
Rebased it now.

stapelberg (approved):

PR #307, opened 2025-03-13
Improve safety of ID allocation

jscissr:
There was an existing mechanism to allocate IDs for sets, but this was using a global counter without any synchronization to prevent data races. I replaced this by a new mechanism which uses a connection-scoped counter, protected by the Conn.mu Mutex. This can then also be used in other places where IDs need to be allocated.

As an additional safeguard, it will panic instead of allocating the same ID twice in a transaction. Most likely, your program will run out of memory before reaching this point.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -17,6 +17,7 @@ package nftables
 import (
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"sync"
 	"syscall"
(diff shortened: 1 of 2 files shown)


stapelberg (approved):

stapelberg:
Thanks!

PR #308, opened 2025-03-13
Add ID to rule

jscissr:
The ID allows referring to a rule before it is committed, as demonstrated in the newly added test.

I had to update all existing tests which compared generated netlink messages against a reference, by inserting the newly added ID attribute.


rule.go
```diff
@@ -30,6 +30,9 @@ const (
 	delRuleHeaderType = netlink.HeaderType((unix.NFNL_SUBSYS_NFTABLES << 8) | unix.NFT_MSG_DELRULE)
 )
 
+// This constant is missing at unix.NFTA_RULE_POSITION_ID.
```
stapelberg, line 33: Can you phrase this in the form of a TODO to include this constant in x/sys/unix please? (Or is there any reason why it wouldn’t be included there?)

rule.go
```diff
@@ -42,10 +45,19 @@ const (
 // A Rule does something with a packet. See also
 // https://wiki.nftables.org/wiki-nftables/index.php/Simple_rule_management
 type Rule struct {
-	Table    *Table
-	Chain    *Chain
+	Table *Table
+	Chain *Chain
+	// Handle identifies an existing Rule.
+	Handle uint64
+	// ID is an identifier for a new Rule, which is assigned by
```
stapelberg, line 52: I’m not 100% clear on the semantics here. This ID is generated by google/nftables, but then also sent to the kernel? Will the kernel use it? Or only users? What is the motivating use-case?

stapelberg (changes requested):

rule.go
jscissr: The way it works is like this:

1. When creating a new rule, google/nftables generates and assigns the ID.
2. Later within the same transaction, the user can delete the rule created in the first step, or insert another rule before or after it. In both cases, the ID is used to identify the rule.
3. During the flush, the kernel can then look up a rule created in an earlier message of the same transaction by its ID.

For example, a transaction could look like this:

- batch begin
- add rule in table T, chain C, with ID 123, and exprs ...
- ...
- delete  rule in table T, chain C, with ID 123
- batch end

Here, a rule is created and later deleted within the same transaction.

Important to note is that the ID is only valid within a single transaction, and cannot be used in a later transaction. This also means that it only needs to be unique within a transaction.

rule.go
jscissr: Done

rule.go
stapelberg: OK, thanks for the explanation!

PR #311, opened 2025-03-26
Expose accumulated message size

nickgarlis:
AFAIU, netlink has a message size limit of ~32KB.

This means that a single transaction would be limited to a few operations. Going over the limit would result in the following error after flushing:

```
netlink receive: recvmsg: no buffer space available
```

If I am not mistaken, the only alternative is to then divide big transactions into smaller chunks. The only issue is that it's hard to tell when the limit has been exceeded.

That's why I suggest exposing a function that would produce the accumulated message size since the `messages` slice is private.

Apologies if there is already another approach to dealing with this. I am still familiarizing myself with this library.   

- https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/netlink/af_netlink.c?id=1e26c5e28ca5821a824e90dd359556f5e9e7b89f#n1930
- https://web.git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=d35c99ff77ecb2eb239731b799386f3b3637a31e

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -329,6 +329,22 @@ func (cc *Conn) FlushRuleset() {
 	})
 }
 
+// GetMessageSize returns the total size of all messages in the buffer.
+// This is useful for making sure that the messages will not exceed the limits
+// of the netlink buffer which is 32768 + ~320 bytes. See also
+// https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/netlink/af_netlink.c?id=1e26c5e28ca5821a824e90dd359556f5e9e7b89f#n1930
+// and
+// https://web.git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=d35c99ff77ecb2eb239731b799386f3b3637a31e
+func (cc *Conn) GetMessageSize() uint32 {
+	cc.mu.Lock()
+	defer cc.mu.Unlock()
+	var total uint32
+	for _, msg := range cc.messages {
+		total += uint32(len(msg.Data)) + 16 // 16 bytes for the header
+	}
+	return total
+}
+
 func (cc *Conn) dialNetlink() (*netlink.Conn, error) {
 	var (
 		conn *netlink.Conn
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -7396,3 +7396,77 @@ func TestSetElementComment(t *testing.T) {
 		}
 	}
 }
+
+func TestGetMessageSize(t *testing.T) {
+	// Use generous socket buffer sizes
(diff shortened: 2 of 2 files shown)


conn.go
```diff
@@ -329,6 +329,22 @@ func (cc *Conn) FlushRuleset() {
 	})
 }
 
+// GetMessageSize returns the total size of all messages in the buffer.
+// This is useful for making sure that the messages will not exceed the limits
+// of the netlink buffer which is 32768 + ~320 bytes. See also
+// https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/netlink/af_netlink.c?id=1e26c5e28ca5821a824e90dd359556f5e9e7b89f#n1930
+// and
+// https://web.git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=d35c99ff77ecb2eb239731b799386f3b3637a31e
+func (cc *Conn) GetMessageSize() uint32 {
```
nickgarlis, line 338: Could always use a different name like `GetBatchSize` or `GetBufferedSize`

stapelberg:
…but this pushes the burden of dealing with the size limit to the user, right?

Why can’t the nftables package transparently deal with the maximum size limit?

nickgarlis:
Seems like it's possible to send "multi-part" messages by utilizing the `Multi` flag in the `Header` and the `Done` flag in the `Type` properties in the message. These correspond to the `NLM_F_MULTI` and `NLMSG_DONE` flags respectively. So, IIUC, it is possible to send chunked messages in a single transaction this way. I will attempt to extend my PR by utilizing those flags in the `Flush` function

- https://github.com/mdlayher/netlink/blob/fbb4dce95f420edbb843b2801504350559a4fa18/message.go#L40
- https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/netlink.h#n63
- https://github.com/mdlayher/netlink/blob/fbb4dce95f420edbb843b2801504350559a4fa18/message.go#L150
- https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/netlink.h#n114

nickgarlis:
Ok, so it seems I was misunderstanding a few things. I had incorrectly assumed that the read buffer being as large as the write buffer should be good enough. Seeing that error made me investigate some other non existent limitation. Turns out that the read buffer should just be apprx. 2-3x the required write buffer size to account for the reply messages and it will work fine. 

I will be closing this PR as it was a user error. Apologies for the confusion.

It could perhaps be nice to have an option that dynamically sets those buffer sizes based on accumulated message size though.

PR #312, opened 2025-04-05
Automatically set socket read & write buffer sizes

nickgarlis:
This is an attempt to port some of the logic that nftables uses to automatically adjust the recvmsg & sndmsg buffer sizes. The implementation of setting sndmsg size is not the same as nftables due to some limitations in the underlying netlink & socket libraries.

You can compare it the behavior to nftables if you:
  - `strace -e trace=sendmsg,setsockopt,getsockopt -s 16384 -xx  ./nftables.test -test.v -run_system_tests -test.run ^TestAutoBufferSize$`
  - `strace -e trace=sendmsg,setsockopt,getsockopt -s 16384 -xx nft --echo -f ./test-rules` where `test-rules` contains https://gist.github.com/nickgarlis/d31cf585b77eef0628118704db8e8310
  
 In my system it's 
```
getsockopt(7, SOL_SOCKET, SO_SNDBUF, [212992], [4]) = 0
setsockopt(7, SOL_SOCKET, SO_SNDBUFFORCE, [458868], 4) = 0
getsockopt(7, SOL_SOCKET, SO_RCVBUF, [212992], [4]) = 0
setsockopt(7, SOL_SOCKET, SO_RCVBUFFORCE, [4197376], 4) = 0 
```
and 
```
getsockopt(3, SOL_SOCKET, SO_SNDBUF, [212992], [4]) = 0
setsockopt(3, SOL_SOCKET, SO_SNDBUF, [2097152], 4) = 0
setsockopt(3, SOL_SOCKET, SO_SNDBUFFORCE, [2097152], 4) = 0
getsockopt(3, SOL_SOCKET, SO_RCVBUF, [212992], [4]) = 0
setsockopt(3, SOL_SOCKET, SO_RCVBUFFORCE, [4197376], 4) = 0
```
respectively.

The change depends on https://github.com/mdlayher/netlink/pull/223

- https://git.netfilter.org/nftables/tree/src/mnl.c?id=713592c6008a8c589a00d3d3d2e49709ff2de62c#n262
- https://git.netfilter.org/libmnl/tree/include/libmnl/libmnl.h?id=03da98bcd284d55212bc79e91dfb63da0ef7b937#n20
- https://git.netfilter.org/nftables/tree/src/mnl.c?id=713592c6008a8c589a00d3d3d2e49709ff2de62c#n391
- https://git.netfilter.org/nftables/tree/src/mnl.c?id=713592c6008a8c589a00d3d3d2e49709ff2de62c#n426

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -250,6 +250,15 @@ func (cc *Conn) Flush() error {
 	}
 	defer func() { _ = closer() }()
 
+	err = cc.setWriteBuffer(conn)
+	if err != nil {
+		return err
+	}
+	err = cc.setReadBuffer(conn)
+	if err != nil {
+		return err
+	}
+
 	messages, err := conn.SendMessages(batch(cc.messages))
 	if err != nil {
 		return fmt.Errorf("SendMessages: %w", err)
@@ -422,4 +431,85 @@ func (cc *Conn) allocateTransactionID() uint32 {
 		cc.lastID = 1
 	}
 	return cc.lastID
+}
+
+// getMessageSize returns the total size of all messages in the buffer.
+func (cc *Conn) getMessageSize() int {
+	var total int
+	for _, msg := range cc.messages {
+		total += len(msg.Data) + 16 // 16 bytes for the header
+	}
+	return total
+}
+
+// setWriteBuffer automatically sets the write buffer of the given connection to
+// the accumulated message size. This is only done if the current write buffer
+// is smaller than the message size.
+//
+// nftables actually handles this differently, it multiplies the number of
+// iovec entries by 2MB. This is not possible to do here as our underlying
+// netlink and socket libraries will only add a single iovec entry and
+// won't expose the number of entries.
+// https://git.netfilter.org/nftables/tree/src/mnl.c?id=713592c6008a8c589a00d3d3d2e49709ff2de62c#n262
+//
+// TODO: Update this function to mimic the behavior of nftables once those
+// limitations are no longer present.
+func (cc *Conn) setWriteBuffer(conn *netlink.Conn) error {
+	messageSize := cc.getMessageSize()
+	writeBuffer, err := conn.WriteBuffer()
+	if err != nil {
+		return err
+	}
+	if writeBuffer < messageSize {
+		return conn.SetWriteBuffer(messageSize)
+	}
+
+	return nil
+}
+
+// getDefaultEchoReadBuffer returns the minimum read buffer size for batches
+// with echo messages.
+//
+// See https://git.netfilter.org/libmnl/tree/include/libmnl/libmnl.h?id=03da98bcd284d55212bc79e91dfb63da0ef7b937#n20
+// and https://git.netfilter.org/nftables/tree/src/mnl.c?id=713592c6008a8c589a00d3d3d2e49709ff2de62c#n391
+func (cc *Conn) getDefaultEchoReadBuffer() int {
+	pageSize := os.Getpagesize()
+	if pageSize < 8192 {
+		return pageSize * 1024
+	}
+
+	return 8192 * 1024
+}
+
+// setReadBuffer  automatically sets the read buffer of the given connection
+// to the required size. This is only done if the current read buffer is smaller
+// than the required size.
+//
+// See https://git.netfilter.org/nftables/tree/src/mnl.c?id=713592c6008a8c589a00d3d3d2e49709ff2de62c#n426
+func (cc *Conn) setReadBuffer(conn *netlink.Conn) error {
+	var bufferSize int
+
+	// If there are any messages with the Echo flag, we initialize the buffer size
+	// to the default echo read buffer size.
+	for _, msg := range cc.messages {
+		if msg.Header.Flags&netlink.Echo == 0 {
+			bufferSize = cc.getDefaultEchoReadBuffer()
+			break
+		}
+	}
+
+	// Just like nftables, we allocate 1024 bytes for each message in the batch.
+	requiredSize := len(cc.messages) * 1024
+	if bufferSize < requiredSize {
+		bufferSize = requiredSize
+	}
+
+	currSize, err := conn.ReadBuffer()
+	if err != nil {
+		return err
+	}
+	if currSize < bufferSize {
+		return conn.SetReadBuffer(bufferSize)
+	}
+	return nil
 }
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -7396,3 +7396,37 @@ func TestSetElementComment(t *testing.T) {
 		}
 	}
 }
+
+func TestAutoBufferSize(t *testing.T) {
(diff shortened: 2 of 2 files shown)


conn.go
```diff
@@ -423,3 +432,84 @@ func (cc *Conn) allocateTransactionID() uint32 {
...
+}
+
+// getDefaultEchoReadBuffer returns the minimum read buffer size for batches
+// with echo messages.
+//
+// See https://git.netfilter.org/libmnl/tree/include/libmnl/libmnl.h?id=03da98bcd284d55212bc79e91dfb63da0ef7b937#n20
+// and https://git.netfilter.org/nftables/tree/src/mnl.c?id=713592c6008a8c589a00d3d3d2e49709ff2de62c#n391
+func (cc *Conn) getDefaultEchoReadBuffer() int {
+	pageSize := os.Getpagesize()
+	if pageSize < 8192 {
```
stapelberg, line 477: This can be simplified to `return max(pageSize, 8192) * 1024`

conn.go
```diff
@@ -423,3 +432,84 @@ func (cc *Conn) allocateTransactionID() uint32 {
...
+// getMessageSize returns the total size of all messages in the buffer.
+func (cc *Conn) getMessageSize() int {
+	var total int
+	for _, msg := range cc.messages {
+		total += len(msg.Data) + 16 // 16 bytes for the header
+	}
+	return total
+}
+
+// setWriteBuffer automatically sets the write buffer of the given connection to
```
stapelberg, line 445: Please rename to enlargeWriteBuffer (with a `set` prefix, I would expect to be able to specify a buffer).

conn.go
```diff
@@ -423,3 +432,84 @@ func (cc *Conn) allocateTransactionID() uint32 {
...
+// is smaller than the message size.
+//
+// nftables actually handles this differently, it multiplies the number of
+// iovec entries by 2MB. This is not possible to do here as our underlying
+// netlink and socket libraries will only add a single iovec entry and
+// won't expose the number of entries.
+// https://git.netfilter.org/nftables/tree/src/mnl.c?id=713592c6008a8c589a00d3d3d2e49709ff2de62c#n262
+//
+// TODO: Update this function to mimic the behavior of nftables once those
+// limitations are no longer present.
```
stapelberg, line 456: Can you replace “those limitations” with something more specific please? It’s not clear to me what you are referring to.

conn.go
```diff
@@ -250,6 +250,15 @@ func (cc *Conn) Flush() error {
 	}
 	defer func() { _ = closer() }()
 
+	err = cc.setWriteBuffer(conn)
+	if err != nil {
```
stapelberg, line 254: nit: here and below, use `if err := cc.enlargeWriteBuffer(conn); err != nil {`

nftables_test.go
```diff
@@ -7396,3 +7396,37 @@ func TestSetElementComment(t *testing.T) {
...
+			Chain: chain,
+			Exprs: []expr.Any{
+				&expr.Verdict{
+					Kind: expr.VerdictAccept,
+				},
+			},
+		})
+	}
+
+	err := conn.Flush()
```
stapelberg, line 7428:
```go
if err := conn.Flush(); err != nil {
```

nftables_test.go
```diff
@@ -7396,3 +7396,37 @@ func TestSetElementComment(t *testing.T) {
...
+		Family: nftables.TableFamilyIPv4,
+		Name:   "test-table",
+	})
+
+	chain := conn.AddChain(&nftables.Chain{
+		Name:  "test-chain",
+		Table: table,
+	})
+
+	for i := 0; i < 4096; i++ {
```
stapelberg, line 7416:
```go
for i := range 4096 {
```

stapelberg (changes requested):
Sorry for the late review. I was checking if the required PR is merged and it isn’t yet, but at this point, might as well review this PR (just in case @mdlayher is waiting for that as a signal…?)

conn.go
nickgarlis: Thanks, I've changed the comment but in case it's not clear enough I am also leaving an explanation here.

If I’m not mistaken, nftables constructs a single netlink message, which is internally split into multiple iovec entries with each being 2MB in size. 

I verified this by using:
```
strace -e trace=setsockopt,getsockopt,sendmsg -s 16384 -xx nft -f ./rules
```
where rules is either:
 - a file that would produce a small message that fits in single 2MB iovec entry
 - a file that would produce a large message that needs multiple iovec entries (the one I attached is not large enough)
 
In both cases, the `sendmsg()` syscall showed a total buffer size (`SO_SNDBUF`) of 2MB per `iovec` entry, suggesting that the message is chunked accordingly.

I think that the socket library used by our netlink implementation doesn't support sending messages composed of multiple iovec segments. As a result, the message must be represented as a single contiguous buffer. That’s why I chose to calculate the full message size in instead.

stapelberg (changes requested):
Looks good overall! I appreciate the detailed comments :) Let me know when you fixed the tests

conn.go
```diff
@@ -423,3 +430,104 @@ func (cc *Conn) allocateTransactionID() uint32 {
...
+	}
+	return total
+}
+
+// canEnlargeBuffers returns true if the connection can automatically enlarge
+// the write and read buffers of the netlink connection.
+func (cc *Conn) canEnlargeBuffers() bool {
+	// If there are sock options, we assume that the user has already set the
+	// buffers to a fixed size.
+	if len(cc.sockOptions) > 0 {
```
nickgarlis, line 448: This is the approach that made most sense to me because it keeps the code backwards compatible. Users will not suddenly get confused when their [explicitly defined buffers](https://github.com/google/nftables/blob/main/nftables_test.go#L7274) change in size.

The only downside I can think of is that if anyone uses `WithSockOptions` for anything other than setting the buffer sizes then they won't benefit from this feature. However, I am not sure it's often or ever the case ?

Anyways, let me know if you have any thoughts on that or an idea for an alternative approach.

nickgarlis:
> Looks good overall! I appreciate the detailed comments :) Let me know when you fixed the tests

@stapelberg I'm glad! Hopefully, it made the reviewing process easier :)

I've now pushed the rest of the changes. [0a7196f](https://github.com/google/nftables/pull/312/commits/0a7196fb65e32d04e4c1c8933c227497fb797f91) fixes the failing tests. 

Thanks again for looking into it!

conn.go
stapelberg: GitHub code search shows 0 usages of nftables.WithSockOptions. 

Can you add a comment to WithSockOptions that explicitly makes users aware of this? Maybe something like:

Note that when using WithSockOptions, you are responsible for providing a large-enough read and write buffer, whereas normally, the nftables package automatically enlarges the buffers as needed.

stapelberg (changes requested):

stapelberg:
Now that the mdlayher/netlink PR was merged, can you update this PR to pick up the new version via go.mod?

PR #313, opened 2025-04-05
Replace %v with %w to wrap underlying errors

corpix:
Main motivation was fix to `ListChain` error to be able to handle `errors.Is(err, unix.ENOENT)`, but then I found more errors which doesn't pass underlying error information and decided to fix them all.

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -215,7 +215,7 @@ func (cc *Conn) ListChain(table *Table, chain string) (*Chain, error) {
 
 	response, err := conn.Execute(msg)
 	if err != nil {
-		return nil, fmt.Errorf("conn.Execute failed: %v", err)
+		return nil, fmt.Errorf("conn.Execute failed: %w", err)
 	}
 
 	if got, want := len(response), 1; got != want {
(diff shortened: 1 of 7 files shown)


stapelberg:
> Main motivation was fix to `ListChain` error to be able to handle `errors.Is(err, unix.ENOENT)`, but then I found more errors which doesn't pass underlying error information and decided to fix them all.

Changing all functions to pass through underlying errors is not “fixing them all”. Whether or not it makes sense to pass through an underlying error needs to be carefully examined on a case-by-case basis. See https://groups.google.com/g/golang-nuts/c/-bYObQTWX_8/m/9iwZrQnIBAAJ and https://matttproud.com/blog/posts/go-errors-and-api-contracts.html for rationale/discussion.

If you want to move forward with your PR, please reduce scope to only the errors you specifically care about and have tested, and then I can take another look. Thanks.

nickgarlis:
@corpix IMO it's only
```golang
fmt.Errorf("conn.Execute failed: %v", err)
```
and 
```golang
fmt.Errorf("receive: %v", err) // used to be "receiveAckAware"
```
that should to wrap the underlying error, since they may carry an errno that can be handled. The rest might not be needed. Were you planning to trim your PR to just those? I’d appreciate this change as I’d like to handle `ENOENT` as you described.

PR #317, opened 2025-06-02
Add optional direction field

glebziz:
# Add optional direction field to connection tracking expressions

This pull request adds support for an optional direction field in ct expressions. The direction field specifies whether the connection tracking information should be retrieved from the original or reply direction of the connection.

## Changes

- Added constants `CtDirOriginal` and `CtDirReply` to represent the possible direction values
- Added a new `OptDirection` boolean field to the `Ct` struct to indicate whether the direction field should be included for certain connection tracking keys
- Modified the marshaling logic to conditionally include the direction field based on the `OptDirection` flag for the following keys ([source](https://wiki.nftables.org/wiki-nftables/index.php/Matching_connection_tracking_stateful_metainformation#ct_packets)):
  - `CtKeyPKTS`
  - `CtKeyBYTES`
  - `CtKeyAVGPKT`
  - `CtKeyL3PROTOCOL`
  - `CtKeyPROTOCOL`
- Fixed the data type for the direction field in marshaling (changed from uint32 to uint8)
- Added logic in the unmarshaling process to detect if a direction was present and set the `OptDirection` flag accordingly
- Added test cases to verify the new functionality

## Motivation

In nftables, certain connection tracking keys like packets and bytes counters can optionally specify a direction. This allows users to retrieve statistics for either the original or reply direction of a connection. Previously, the library didn't support specifying a direction for these keys, limiting the flexibility of connection tracking expressions.

## Backward Compatibility

This change is fully backward compatible:
- Existing code that doesn't use the direction field for these keys will continue to work as before
- The `OptDirection` field defaults to false, maintaining the previous behavior
- The unmarshaling process correctly handles both cases (with and without direction)

## Example Usage

```go
// Create a ct expression to get packet count from the original direction
ct := &expr.Ct{
    Register:     1,
    Key:          expr.CtKeyPKTS,
    Direction:    expr.CtDirOriginal,
    OptDirection: true,
}

// Create a ct expression to get packet count without specifying direction
ct := &expr.Ct{
    Register: 1,
    Key:      expr.CtKeyPKTS,
}
```

### DIFF
--- a/expr/ct.go
+++ b/expr/ct.go
@@ -110,6 +110,12 @@ const (
 	CtStateUDPREPLIED
 )
 
+const (
+	// https://git.netfilter.org/libnftnl/tree/src/expr/ct.c?id=116e95aa7b6358c917de8c69f6f173874030b46b#n31
+	CtDirOriginal = iota
+	CtDirReply
+)
+
 // https://git.netfilter.org/libnftnl/tree/src/obj/ct_timeout.c?id=116e95aa7b6358c917de8c69f6f173874030b46b#n57
 var CtStateUDPTimeoutDefaults CtStatePolicyTimeout = map[uint16]uint32{
 	CtStateUDPUNREPLIED: 30,
@@ -122,6 +128,7 @@ type Ct struct {
 	SourceRegister bool
 	Key            CtKey
 	Direction      uint32
+	OptDirection   bool
 }
 
 func (e *Ct) marshal(fam byte) ([]byte, error) {
@@ -165,10 +172,16 @@ func (e *Ct) marshalData(fam byte) ([]byte, error) {
 	exprData = append(exprData, regData...)
 
 	switch e.Key {
+	case CtKeyPKTS, CtKeyBYTES, CtKeyAVGPKT, CtKeyL3PROTOCOL, CtKeyPROTOCOL:
+		if !e.OptDirection {
+			break
+		}
+
+		fallthrough
 	case CtKeySRC, CtKeyDST, CtKeyPROTOSRC, CtKeyPROTODST, CtKeySRCIP, CtKeyDSTIP, CtKeySRCIP6, CtKeyDSTIP6:
 		regData, err = netlink.MarshalAttributes(
 			[]netlink.Attribute{
-				{Type: unix.NFTA_CT_DIRECTION, Data: binaryutil.BigEndian.PutUint32(e.Direction)},
+				{Type: unix.NFTA_CT_DIRECTION, Data: []byte{uint8(e.Direction)}},
 			},
 		)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


glebziz:
Hi @stapelberg! This PR addresses two issues with connection tracking:

1. **Fixed Direction Field Size**: PR fixes the size of the direction field in the `Ct` structure. Previously it used `uint32` internally, but in nftables this field is of type [`uint8`](https://git.netfilter.org/libnftnl/tree/src/expr/ct.c?id=116e95aa7b6358c917de8c69f6f173874030b46b#n28), which resulted in the error `netlink: attribute 0x3 is not a uint32; length: 1`.

2. **Added Optional Direction Support**: Previously, the `Direction` field was ignored when connection tracking had certain key types like `CtKeyPKTS`. This meant that code like the following would generate a rule without the "original" direction specifier:
```go
expr.Ct{
    Register:     1,
    Key:          expr.CtKeyPKTS,
    Direction:    expr.CtDirOriginal, // This was ignored for packet metadata keys
},
```

```nftables
iifname "br-lan" ip daddr @ip_set ct packets < 8 queue to 200
```

The library now supports an optional direction field in connection tracking expressions through the new `OptDirection` field. This enhancement allows you to specify a direction (original/reply) for packet metadata keys like `CtKeyPKTS`, `CtKeyBYTES`, etc.

For example, you can now create rules like:
```
iifname "br-lan" ip daddr @ip_set ct original packets < 8 queue to 200
```

Using the following Go code:
```go
rule := &nftables.Rule{
    Table: table,
    Chain: postnat,
    Exprs: []expr.Any{
        &expr.Meta{
            Key:      expr.MetaKeyIIFNAME,
            Register: 1,
        },
        &expr.Cmp{
            Op:       expr.CmpOpEq,
            Register: 1,
            Data:     []byte("br-lan\x00"),
        },
        &expr.Meta{
            Key:      expr.MetaKeyNFPROTO,
            Register: 1,
        },
        &expr.Cmp{
            Op:       expr.CmpOpEq,
            Register: 1,
            Data:     []byte{2},
        },
        &expr.Payload{
            DestRegister: 1,
            Base:         expr.PayloadBaseNetworkHeader,
            Offset:       16,
            Len:          4,
        },
        &expr.Lookup{
            SourceRegister: 1,
            SetName:        ipSet.Name,
            SetID:          ipSet.ID,
        },
        &expr.Ct{
            Register:     1,
            Key:          expr.CtKeyPKTS,
            Direction:    expr.CtDirOriginal,
            OptDirection: true,  // This flag enables the direction field for packet metadata keys
        },
        &expr.Cmp{
            Op:       expr.CmpOpLt,
            Register: 1,
            Data:     []byte{8, 0, 0, 0, 0, 0, 0, 0},
        },
        &expr.Queue{
            Num:  200,
        },
    },
}
```

stapelberg:
Thank you!

PR #319, opened 2025-07-09
fix set.KeyType ovewrite for vmap

NoobsEnslaver:
Just add vmap by hands like
```shell
nft add table inet filter
nft add chain inet filter input
nft add map inet filter ipmap { type ipv4_addr : verdict; }
```
then check output of `nftables.(*Conn).GetSetByName`, will be `verdict:incorrect` map, but `ipv4_addr:verdict` expected. Now fixed.


stapelberg:
Thanks for your PR! Would it be possible to add a test? Ideally, the test would fail without your fix, and pass with your fix. That way, we can make sure the functionality doesn’t break with future changes…

PR #322, opened 2025-08-07
add DelTable(force)

NoobsEnslaver:
add `nft destroy table ...` functional

### DIFF
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -3818,7 +3818,7 @@ func TestDeleteElementNamedSet(t *testing.T) {
 		Name:    "test",
 		KeyType: nftables.TypeInetService,
 	}
-	if err := c.AddSet(portSet, []nftables.SetElement{{Key: []byte{0, 22}}, {Key: []byte{0, 23}}}); err != nil {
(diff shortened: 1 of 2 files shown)


table.go
```diff
@@ -50,16 +55,23 @@ type Table struct {
...
-func (cc *Conn) DelTable(t *Table) {
+func (cc *Conn) DelTable(t *Table, force ...bool) {
 	cc.mu.Lock()
 	defer cc.mu.Unlock()
 	data := cc.marshalAttr([]netlink.Attribute{
 		{Type: unix.NFTA_TABLE_NAME, Data: []byte(t.Name + "\x00")},
 		{Type: unix.NFTA_TABLE_FLAGS, Data: []byte{0, 0, 0, 0}},
 	})
+
+	var hdrType netlink.HeaderType
```
stapelberg, line 66: nit: I would prefer expressing this code as follows:

```go
	hdrType := delTableHeaderType
	if len(force) > 0 && force[0] {
		hdrType = destroyTableHeaderType
	}
```

This way, we initialize the variable to its default (a little clearer), and we only consider the first bool passed to the function.

(But maybe this comment is moot after addressing my other comment.)

stapelberg (changes requested):

Issue #323, opened 2025-08-11
Transaction does not fail when adding an invalid set


nickgarlis:
## Problem
When adding an invalid set (e.g., an anonymous set that is not constant), the `AddSet` call returns an error as expected.
However, the subsequent `Flush()` does not fail, and the transaction still partially succeeds.
This results in an inconsistent state where the invalid set is missing but other objects (e.g., the table) remain created.

## Expected behavior
If a transaction contains an invalid set, the entire transaction should fail during `Flush()` (or earlier), ensuring atomicity.

## Reproduction code

```go
func TestInvalidSetError(t *testing.T) {
	conn, newNS := nftest.OpenSystemConn(t, *enableSysTests)
	defer nftest.CleanupSystemConn(t, newNS)
	conn.FlushRuleset()
	defer conn.FlushRuleset()

	table := conn.AddTable(&nftables.Table{
		Family: nftables.TableFamilyIPv4,
		Name:   "test-table",
	})

	// Create an anonymous set that is not constant to trigger an error
	set := &nftables.Set{
		Table:     table,
		KeyType:   nftables.TypeIPAddr,
		Anonymous: true,
	}
	err := conn.AddSet(set, []nftables.SetElement{})
	if err == nil {
		t.Error("expected error when adding an anonymous that is not constant")
	}

	// The messages slice is now incomplete. It is missing the invalid set.
	// This makes it so that the flush succeeds, but the set is not created.
	if err := conn.Flush(); err == nil {
		t.Errorf("expected error when flushing rules with an invalid set, got nil")
	}

	createdTable, err := conn.ListTable(table.Name)
	if createdTable != nil || err == nil {
		t.Errorf("expected table to not exist, got %v, error: %v", createdTable, err)
	}
}
```

### Accumulate errors until flush
  - Update/create/delete operations would not immediately return an error.
  - Instead, they would append any encountered errors to an internal slice.
  - When `Flush()` is called, it would check this slice. If errors are present, it clears the [messages](https://github.com/google/nftables/blob/508bb1ffd4a2dc62c6c5b59152ff94e5ab9dc22d/conn.go#L44) slice and short-circuits the commit.

This approach would require `AddSet` (and similar functions) to stop returning errors ( I don't think there are many such instances). It could also create an opportunity to add richer, more descriptive errors to make the library easier to debug. For example, returning a specific error when `DelSet` is called on an anonymous set (an invalid operation). 

Some considerations:
[...]

PR #324, opened 2025-08-18
Defer errors until Flush to avoid incomplete batches

nickgarlis:
Any create/update/delete operation that returns a validation or marshalling error can leave the message batch in an incomplete state due to short-circuiting. This can result in either:

  - Non-atomic transactions if Flush is called (incomplete batch)
  - Users being unable to clear the incomplete batch (no API exposed)

This change ensures that errors are collected and deferred until Flush. Instead of returning immediately, the following methods now append errors to a slice checked at Flush:

  - AddSet
  - DelRule
  - SetAddElements

See: https://github.com/google/nftables/issues/323

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -42,7 +42,7 @@ type Conn struct {
 	lasting      bool       // establish a lasting connection to be used across multiple netlink operations.
 	mu           sync.Mutex // protects the following state
 	messages     []netlinkMessage
-	err          error
+	errs         []error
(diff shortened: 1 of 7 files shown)


NoobsEnslaver:
```
Users being unable to clear the incomplete batch (no API exposed)
```
just close connection without flush?

nickgarlis:
> just close connection without flush? 

Yeah that could be an option but IIUC, that's not something we want to do when the connection is lasting.

PR #326, opened 2025-09-09
Add GetRuleByHandle, ResetRule & ResetRules methods

nickgarlis:
This commit introduces the following methods:

  - GetRuleByHandle
  - ResetRule
  - ResetRules

It also refactors GetRules and the deprecated GetRule methods to share a common getRules implementation.

### DIFF
--- a/rule.go
+++ b/rule.go
@@ -71,31 +71,98 @@ type Rule struct {
 
 // GetRule returns the rules in the specified table and chain.
 //
-// Deprecated: use GetRules instead.
+// Deprecated: use GetRuleByHandle instead.
 func (cc *Conn) GetRule(t *Table, c *Chain) ([]*Rule, error) {
 	return cc.GetRules(t, c)
 }
 
+// GetRuleByHandle returns the rule in the specified table and chain by its
+// handle.
+// https://docs.kernel.org/networking/netlink_spec/nftables.html#getrule
+func (cc *Conn) GetRuleByHandle(t *Table, c *Chain, handle uint64) (*Rule, error) {
+	rules, err := cc.getRules(t, c, unix.NFT_MSG_GETRULE, handle)
+	if err != nil {
+		return nil, err
+	}
+
+	if got, want := len(rules), 1; got != want {
+		return nil, fmt.Errorf("expected rule count %d, got %d", want, got)
+	}
+
+	return rules[0], nil
+}
+
 // GetRules returns the rules in the specified table and chain.
 func (cc *Conn) GetRules(t *Table, c *Chain) ([]*Rule, error) {
+	return cc.getRules(t, c, unix.NFT_MSG_GETRULE, 0)
+}
+
+// ResetRule resets the stateful expressions (e.g., counters) of the given
+// rule. The reset is applied immediately (no Flush is required). The returned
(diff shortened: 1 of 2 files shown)


rule.go
```diff
@@ -71,31 +71,98 @@ type Rule struct {
...
-// Deprecated: use GetRules instead.
+// Deprecated: use GetRuleByHandle instead.
 func (cc *Conn) GetRule(t *Table, c *Chain) ([]*Rule, error) {
 	return cc.GetRules(t, c)
 }
 
+// GetRuleByHandle returns the rule in the specified table and chain by its
+// handle.
+// https://docs.kernel.org/networking/netlink_spec/nftables.html#getrule
+func (cc *Conn) GetRuleByHandle(t *Table, c *Chain, handle uint64) (*Rule, error) {
```
nickgarlis, line 82: Here I assumed that we want to keep consistency with the rest of the rule methods in the file. It could have been `(c *Chain, handle uint64)` instead but I don't it'd be clear to the user that the Table pointer cannot be nil.

rule.go
```diff
@@ -71,31 +71,98 @@ type Rule struct {
...
 func (cc *Conn) GetRules(t *Table, c *Chain) ([]*Rule, error) {
+	return cc.getRules(t, c, unix.NFT_MSG_GETRULE, 0)
+}
+
+// ResetRule resets the stateful expressions (e.g., counters) of the given
+// rule. The reset is applied immediately (no Flush is required). The returned
+// rule reflects its state prior to the reset. The provided rule must have a
+// valid Handle.
+// https://docs.kernel.org/networking/netlink_spec/nftables.html#getrule-reset
+func (cc *Conn) ResetRule(t *Table, c *Chain, handle uint64) (*Rule, error) {
```
nickgarlis, line 105: Same assumptions as above but this could perhaps be (r *Rule) instead.

rule.go
```diff
@@ -71,31 +71,98 @@ type Rule struct {
...
+	if got, want := len(rules), 1; got != want {
+		return nil, fmt.Errorf("expected rule count %d, got %d", want, got)
+	}
+
+	return rules[0], nil
+}
+
+// ResetRules resets the stateful expressions (e.g., counters) of all rules
+// in the given table and chain. The reset is applied immediately (no Flush
+// is required). The returned rules reflect their state prior to the reset.
```
stapelberg, line 124: Huh, it’s a little surprising to read “no Flush is required”. Is this documented upstream somewhere?

rule.go
stapelberg: > It could have been `(c *Chain, handle uint64)` instead but I don't think it'd be clear to the user that the Table pointer cannot be nil.

I’m having trouble parsing this sentence. The code looks like `*Table` is required, and you say `the Table pointer cannot be nil`, so then how could this signature possibly be `(c *Chain, handle uint64)`  (no Table pointer)…?

stapelberg (changes requested):

rule.go
nickgarlis: Sorry I meant the *Table pointer inside of the chain object. I am not sure whether users would expect &Chain{Name: "test-chain} to be an invalid input.

PR #328, opened 2025-09-24
Accumulate errors until Flush to avoid incomplete batches

nickgarlis:
Any create/update/delete (batched) operation that returns a validation or marshalling error could leave the message batch in an incomplete state due to short-circuiting. This leads to two issues:

  - Non-atomic transactions if Flush is called (incomplete batch)
  - No way for users to clear the incomplete batch without sending it to the kernel

With this change, errors are accumulated and returned during Flush as well.

See: https://github.com/google/nftables/issues/323


stapelberg:
Thank you!

PR #334, opened 2025-11-03
Refactor receive logic to reuse message reading sequence

nickgarlis:
Introduce a `receiveSeq` iterator to simplify reading messages from the socket without copying them into a temporary slice. It yields only messages relevant to nftables operations (i.e., those containing nftables attributes).

Also refactor `receiveAckAware` into a reusable `receive` method, which, like `flush`, uses the same `receiveSeq` iterator for consistency.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -17,6 +17,7 @@ package nftables
 import (
 	"errors"
 	"fmt"
+	"iter"
 	"math"
 	"os"
 	"sync"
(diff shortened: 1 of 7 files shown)


conn.go
```diff
@@ -156,34 +157,86 @@ func (cc *Conn) netlinkConnUnderLock() (*netlink.Conn, netlinkCloser, error) {
...
-		return nil, err
+// receive will drain the receive buffer of the provided netlink connection
+// and return all received messages, along with the first error encountered,
+// if any.
+func (cc *Conn) receive(conn *netlink.Conn) ([]netlink.Message, error) {
+	var allReplies []netlink.Message
+	var firstError error
+
+	for msg, err := range cc.receiveSeq(conn) {
+		if err != nil && firstError == nil {
```
nickgarlis: receiveAckAware returns the first receive error, flush joins all receive errors. Perhaps we could join them here too for consistency

stapelberg: I’m in favor of consistency, but I wonder if we should go the other way and make Flush also only return the first error? I’m not sure how useful the non-first errors are…

PR #335, opened 2025-11-06
Enhance error tracing

nickgarlis:
Associate netlink errors with the specific NFT operations that produced them.

- Add nftMsgType with a String() method to replace sys/unix constants
- Add handleReceiveError to match errors to their originating messages
- Add withOpError to assist in handling netlink.OpError values

This updates errors from:

    receive: netlink receive: no such file or directory

to:

    receive: NFT_MSG_NEWCHAIN: netlink receive: no such file or directory


stapelberg (approved):

stapelberg:
Thanks!

PR #336, opened 2025-11-07
Stop requesting ACKs

nickgarlis:
Since https://github.com/google/nftables/pull/330 and https://github.com/google/nftables/pull/334 were merged, I wonder whether we need to keep requesting an acknowledge for every message we send.

Skipping ACK requests improves runtime and memory usage for large Flush batches, as it reduces the number of required `recvmsg` calls.

The first error will still be reported via an acknowledge even if none are requested.

This also matches the behavior of nft userspace.

To validate this, I prepared a small stress test that inserts (math.MaxUint16 * 10) rules and then calls Flush. The test completed in 13.36s with ACKs enabled, versus 2.98s without ACKs.

A separate memory test focusing on Flush also showed noticeably lower usage when ACKs were not requested (see attached images).

2025-11-07_with_acks
2025-11-07_without_acks

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -143,7 +143,7 @@ func (cc *Conn) AddChain(c *Chain) *Chain {
 	cc.messages = append(cc.messages, netlinkMessage{
 		Header: netlink.Header{
 			Type:  nftMsgNewChain.HeaderType(),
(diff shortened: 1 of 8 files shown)


nickgarlis:
Thanks! I've rebased now

Issue #247, opened 2023-11-09
Adding rules in code produces different results and logs than the rules I added directly from the command line


xlango:
When i used command lines 
{ 
	nft add table ip filter
	nft add chain ip filter output { type filter hook output priority 0\; }
	nft add set  filter ipSet { type ipv4_addr \; flags interval\;}
	nft add element ip filter ipSet {10.34.11.179}
	nft add set  filter portSet { type inet_service \; flags interval\;}
	nft add element ip filter portSet {1234}
	nft add rule ip filter output ip daddr @ipSet tcp dport @portSet counter log drop	
}

to add a rule related to set, it worked correctly.
But when i used nftables-main to add a similar rule, it blocked tcp flow to "127.0.0.1" too.
The code is :
 func main()  {
	c, err := nftables.New()
	if err != nil {
		return
	}

	c.FlushRuleset()

	filter := c.AddTable(&nftables.Table{
		Family: nftables.TableFamilyIPv4,
		Name:   "filter",
	})

	input := c.AddChain(&nftables.Chain{
		Name:     "output",
		Hooknum:  nftables.ChainHookOutput,
		Priority: nftables.ChainPriorityFilter,
		Table:    filter,
		Type:     nftables.ChainTypeFilter,
	})
	ipSet := &nftables.Set{
		Name:"ipSet",
		Table:     filter,
		Interval:      true,
		Concatenation: true,
		KeyType:   nftables.TypeIPAddr,
	}
	if err := c.AddSet(ipSet, []nftables.SetElement{
		{
			Key: []byte(net.ParseIP("10.34.11.179").To4()),
			KeyEnd: []byte(net.ParseIP("10.34.11.180").To4()),
		},
	}); err != nil {
		return
	}

	portSet := &nftables.Set{
		Name:"portSet",
		Table:     filter,
		Interval:      true,
		Concatenation: true,
		KeyType:    nftables.TypeInetService,
	}

	if err := c.AddSet(portSet, []nftables.SetElement{
		{
			Key: binaryutil.BigEndian.PutUint16(1234),
			KeyEnd:binaryutil.BigEndian.PutUint16(1235),
		},
	}); err != nil {
		return
	}

	c.AddRule(&nftables.Rule{
		Table: filter,
		Chain: input,
		Exprs: []expr.Any{
			&expr.Payload{
				DestRegister: 1,
				Base:         expr.PayloadBaseNetworkHeader,
				Offset:       16,
				Len:          4,
			},

[...]

PR #342, opened 2025-11-15
Add NetInterval helper

nickgarlis:
When creating set elements that represent a network, the interval range must be half-open [start, end) rather than inclusive [start, end]. For example, for 10.0.0.0/24, the expected range is 10.0.0.0 to 10.0.1.0 instead of 10.0.0.0 to 10.0.0.255.

This change introduces a NetInterval helper that returns the correct range given a CIDR string.

Some notes for consideration:
  - I used `net.IP` for consistency with `NetFirstAndLastIP`. However, `netip.Addr` seems to be the new standard. Should I change this PR to convert the returned values of `NetInterval` to `netip.Addr` ?
  - Would this package benefit from a dependency like https://github.com/go4org/netipx so that those utility functions don't need to be re-implemented ?

### DIFF
--- a/util.go
+++ b/util.go
@@ -87,3 +87,42 @@ func NetFirstAndLastIP(networkCIDR string) (first, last net.IP, err error) {
 
 	return first, last, nil
 }
+
+// nextIp returns the next IP address after the given one.
+// If the next address overflows, the sentinel values 0.0.0.0 (IPv4)
+// or :: (IPv6) are returned.
+func nextIP(ip net.IP) net.IP {
+	if ip == nil {
+		return nil
+	}
+
+	next := make(net.IP, len(ip))
+	copy(next, ip)
+
+	for i := len(next) - 1; i >= 0; i-- {
+		next[i]++
+		if next[i] != 0 {
+			return next
+		}
+	}
+
+	// All bytes overflowed to 0
+	return next
+}
+
+// NetInterval returns the half-open ([start, end)) interval of a CIDR string.
+// This is the range that nftables uses for interval matching with set elements.
+// Unlike NetFirstAndLastIP, the end value is one past the last IP in the
+// network. If the last IP is overflowed, the end value will be a zero IP.
+//
+// For example, for the CIDR "10.0.0.0/24", NetInterval returns
+// first=10.0.0.0 and last=10.0.1.0. Note that last is one more than the
+// broadcast address of the CIDR.
+func NetInterval(cidr string) (net.IP, net.IP, error) {
+	first, last, err := NetFirstAndLastIP(cidr)
+	if err != nil {
+		return first, last, err
+	}
+
+	return first, nextIP(last), nil
+}
--- a/util_test.go
+++ b/util_test.go
@@ -76,3 +76,128 @@ func TestNetFirstAndLastIP(t *testing.T) {
 		})
 	}
 }
+
+func TestNetInterval(t *testing.T) {
+	tests := []struct {
+		name        string
+		cidr        string
+		wantFirstIP net.IP
+		wantLastIP  net.IP
+		wantErr     bool
+	}{
+		{
+			name:        "Test Invalid",
+			cidr:        "invalid-cidr",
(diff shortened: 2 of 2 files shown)


stapelberg:
> // This is the range that nftables uses for interval matching with set elements.

Can you explain when/where/how nftables uses such an interval? I would also like to add a reference to the wraparound behavior (ending up with :: or 0.0.0.0), because that sounds a little unexpected…

> * I used `net.IP` for consistency with `NetFirstAndLastIP`. However, `netip.Addr` seems to be the new standard. Should I change this PR to convert the returned values of `NetInterval` to `netip.Addr` ?

I haven’t worked much with the new netip type. I would hold off on making this change for now. We can consider updating all net.IP to netip.Addr independently.

> * Would this package benefit from a dependency like https://github.com/go4org/netipx so that those utility functions don't need to be re-implemented ?

No, I would like to keep the dependencies as minimal as possible for this project.

nickgarlis:
> Can you explain when/where/how nftables uses such an interval?

When you add a CIDR element to an nftables set, nft internally represents that CIDR as two separate interval elements.

For example, adding `10.0.0.0/24` results in the following elements:

```golang
[]nftables.SetElement{
	{Key: net.ParseIP("10.0.0.0").To4()},
	{Key: net.ParseIP("10.0.1.0").To4(), IntervalEnd: true},
}
```
You can see this by running: 
 ```
 nft --debug=netlink add element ip test-table test-set { 10.0.0.0/24 }
 ``` 
which prints something like:
```
test-set test-table 0
	element 0000000a  : 0 [end]	element 0001000a  : 1 [end]
```
If the set is initially empty, nft may show an extra unrelated element. That behavior is explained in more detail in #247 and is outside the scope of this PR.

## What happens if you use the normal range?

If you instead create the elements like this:
```golang
[]nftables.SetElement{
	{Key: net.ParseIP("10.0.0.0").To4()},
	{Key: net.ParseIP("10.0.0.255").To4(), IntervalEnd: true},
}
```
and then inspect the set with `nft`, you’ll see:
```
table ip test-table {
	set test-set {
		type ipv4_addr
		flags interval
		elements = { 10.0.0.0-10.0.0.254 }
	}
}
```
This shows that nft normalizes interval boundaries itself so the correct approach should be to follow nft’s internal interval representation.

## Edge case: ranges that overflow

If the final IP in the interval overflows (e.g., `255.255.255.254/31`) , nft expresses this using a single element and a userdata flag indicating that no end element exists:

```
test-set test-table 0
	element feffffff  : 0 [end]  userdata = { \x01\x04\x01\x00\x00\x00 }
```
With #343 this would be addressed like this:
```golang
 []nftables.SetElement{
	{Key: net.ParseIP("255.255.255.254").To4(), IntervalOpen: true},
}
```

## Putting it all together

Given an arbitrary CIDR, this is how I would use those two features together:

```golang
var elements []nftables.SetElement
first, last, err := nftables.NetInterval(unknownCIDR)
if err != nil {
	return nil, err
}
// last overflowed - there is no end element
if last.IsUnspecified() { // perhaps last == nil would be better ?
	elements = []nftables.SetElement{
		{
			Key:          first,
			IntervalOpen: true,
		},
	}
} else {
	elements = []nftables.SetElement{
		{
			Key: first,
		},
		{
			Key:         last,
			IntervalEnd: true,
		},
	}
}
```

Please let me know if you would like me to clarify anything. I am happy to go into more detail.

> I would also like to add a reference to the wraparound behavior (ending up with :: or 0.0.0.0), because that sounds a little unexpected…

I think the explanation above covers this, but just to make sure we are aligned: the zero IP is used to represent "no end IP", since in some cases, the last IP would go past the maximum valid address. Would using a pointer be a better approach ?

> I haven’t worked much with the new netip type. I would hold off on making this change for now. We can consider updating all net.IP to netip.Addr independently.

Sounds good :+1: 

> No, I would like to keep the dependencies as minimal as possible for this project.

Also sounds good :+1:

tacerus:
Thank you for this - very useful!

Issue #346, opened 2025-12-05
Merge set elements


tacerus:
Hi,

`nft` will print intervals in CIDR representation - for example:

```
inet filter @testset6
        element 000080fe 00000000 00000000 02000000  : 1 [end]
        element 000080fe 00000000 00000000 01000000  : 0 [end]
        element 0000072a 01000000 00000000 00000000  : 1 [end]
        element 0000072a 00000000 00000000 00000000  : 0 [end]
        element 00000000 00000000 00000000 00000000  : 1 [end]
inet filter @testset4
        element 0001a8c0  : 1 [end]
        element 0000a8c0  : 0 [end]
        element 0200007f  : 1 [end]
        element 0100007f  : 0 [end]
        element 0200000a  : 1 [end]
        element 0100000a  : 0 [end]
        element 00000000  : 1 [end]
```

gets turned into:

```
table inet filter {
        set testset6 {
                type ipv6_addr
                flags interval
                elements = { 2a07::/64,
                             fe80::1 }
        }
        set testset4 {
                type ipv4_addr
                flags interval
                elements = { 10.0.0.1, 127.0.0.1,
                             192.168.0.0/24 }
        }
}
```

Thanks to https://github.com/google/nftables/pull/342 it is now easy to add new elements in this format.

I would appreciate some advice in regards to retrieving elements in the same representation.

One issue is https://github.com/google/nftables/issues/320, but it can be worked around by iterating over the returned elements in reverse order (hoping the order is stable).

So far I came up with the following (omitted set discovery and error handling for brevity):

```
        elements, _ := nft.GetSetElements(set)

        var (
                out []string  // will contain the resulting nft-style elements
                last nftables.SetElement
        )

        NullIPv4 := []byte{0, 0, 0, 0}
        NullIPv6 := []byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}

        for i := len(elements)-1; i >= 0; i-- {
                e := elements[i]
                switch set.KeyType.Name {
                case "ipv4_addr", "ipv6_addr":
                        if bytes.Compare(e.Key, NullIPv4) == 0 || bytes.Compare(e.Key, NullIPv6) == 0 {
                                continue
                        }

                        ipCurrent, _ := netip.AddrFromSlice(e.Key)
                        ipLast, _ := netip.AddrFromSlice(last.Key)
                        ipLastNext := ipLast.Next()

[...]

PR #347, opened 2025-12-06
Add NetFromRange and NetFromInterval helpers

tacerus:
When retrieving set elements it can be desired to format the result in the same way `nft` would, which is merging intervals to CIDR representations. To make this easier, introduce helper functions which allow for conversion of IP address ranges to CIDR networks.

https://github.com/google/nftables/issues/346

### DIFF
--- a/util.go
+++ b/util.go
@@ -16,10 +16,17 @@ package nftables
 
 import (
 	"encoding/binary"
+	"errors"
 	"net"
+	"net/netip"
 
 	"github.com/google/nftables/binaryutil"
 	"golang.org/x/sys/unix"
+)
+
+var (
+	MaxIPv4 = net.IP{255, 255, 255, 255}
+	MaxIPv6 = net.IP{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff}
 )
 
 func extraHeader(family uint8, resID uint16) []byte {
@@ -125,4 +132,84 @@ func NetInterval(cidr string) (net.IP, net.IP, error) {
 	}
 
 	return first, nextIP(last), nil
+}
+
+// endIp returns the last address in a given network.
+func endIp(netIp net.IP, mask net.IPMask) net.IP {
+	ip := make(net.IP, len(netIp))
+	copy(ip, netIp)
+
+	for i := 0; i < len(mask); i++ {
+		ipIdx := len(ip) - i - 1
+		ip[ipIdx] = netIp[ipIdx] | ^mask[len(mask)-i-1]
+	}
+
+	return ip
+}
+
+// NetFromRange returns a CIDR IP network given a start and end address.
+// If the network is an exact match, ok will be true.
+func NetFromRange(first net.IP, last net.IP) (*net.IPNet, bool, error) {
+	ip1 := net.IP(first)
+	ip2 := net.IP(last)
+
+	maxLen := 32
+	isIpv6 := ip1.To4() == nil
+
+	if isIpv6 && ip2.To4() != nil || !isIpv6 && ip2.To4() == nil {
+		return nil, false, errors.New("Cannot mix IPv4 and IPv6 or process empty IP.")
+	}
+
+	if isIpv6 {
+		maxLen = 128
+	}
+
+	var match *net.IPNet
+	for l := maxLen; l >= -1; l-- {
+		cidrmask := net.CIDRMask(l, maxLen)
+		ipmask := ip2.Mask(cidrmask)
+		ipnet := net.IPNet{
+			IP:   ipmask,
+			Mask: cidrmask,
+		}
+
+		if ipnet.Contains(ip1) {
+			match = &ipnet
+			break
+		}
+
+	}
+
+	matchFirst := match.IP.Equal(ip1)
+
+	// short-circuit if first address is not start of the network
+	if !matchFirst {
+		return match, matchFirst, nil
+	}
+
+	return match, endIp(match.IP, match.Mask).Equal(ip2), nil
+}
+
+// NetFromInterval returns a CIDR IP network given a start and end address as found in intervals.
+// This is similar to NetFromRange, but subtracts one address from the end of the range.
+// If the resulting network is an exact match, ok will be true.
+func NetFromInterval(first net.IP, last net.IP) (out *net.IPNet, ok bool, err error) {
+	var previous net.IP
+
+	if len(last) == 0 {
+		if first.To4() == nil {
+			previous = MaxIPv6
+		} else {
+			previous = MaxIPv4
+		}
+	} else {
+		ip2, ok := netip.AddrFromSlice(last)
+		if !ok {
+			return nil, false, errors.New("Failed to construct slice from network.")
+		}
+
+		previous = ip2.Prev().AsSlice()
+	}
+
+	return NetFromRange(first, previous)
 }
--- a/util.go
+++ b/util.go
@@ -148,7 +148,7 @@ func endIp(netIp net.IP, mask net.IPMask) net.IP {
 }
 
 // NetFromRange returns a CIDR IP network given a start and end address.
-// If the network is an exact match, ok will be true.
+// If an exact match is found, ok will be true. If not, no IPNet will be returned, and ok will be false.
 func NetFromRange(first net.IP, last net.IP) (*net.IPNet, bool, error) {
 	ip1 := net.IP(first)
 	ip2 := net.IP(last)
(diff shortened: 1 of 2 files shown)


tacerus:
When using this for the case described in https://github.com/google/nftables/issues/346, it also needs to be considered that `nft` will not print a CIDR mask if it is 32 / 128 - I'm not sure this should be considered in the functions introduced here (which I think would require having the functions return something other than `net.IPNet`s) or in later logic.

Edit: I am thinking to add another function like `FormatIpNet` which takes an IPNet as input, and returns a string (either with or without the mask, depending on how nft would treat it).

util.go
```diff
@@ -126,3 +129,58 @@ func NetInterval(cidr string) (net.IP, net.IP, error) {
...
+	_, out, err := net.ParseCIDR(fmt.Sprintf("%s/%d", address, cidr))
+	if err != nil {
+		return nil, err
+	}
+
+	return out, err
+}
+
+// NetFromRange returns a CIDR IP network given a start and end address
+func NetFromRange(first []byte, last []byte) (*net.IPNet, error) {
```
nickgarlis: I wonder whether it'd be best for first, last to be net.IP instead.

util.go
```diff
@@ -126,3 +129,58 @@ func NetInterval(cidr string) (net.IP, net.IP, error) {
...
+
+// NetFromNetInterval returns a CIDR IP network given a start and end address as found in intervals.
+// This is similar to NetFromRange, but subtracts one address from the end of the range.
+func NetFromIntervalRange(first []byte, last []byte) (out *net.IPNet, err error) {
+	ip2, ok := netip.AddrFromSlice(last)
+	if !ok {
+		return nil, errors.New("Failed to construct slice from network.")
+	}
+
+	previous := ip2.Prev()
```
nickgarlis: What happens if there is no previous IP ?

util.go
```diff
@@ -126,3 +129,58 @@ func NetInterval(cidr string) (net.IP, net.IP, error) {
...
+	for l := maxLen; l >= 0; l-- {
+		cidrmask := net.CIDRMask(l, maxLen)
+		ipmask := ip2.Mask(cidrmask)
+		ipnet := net.IPNet{
+			IP:   ipmask,
+			Mask: cidrmask,
+		}
+
+		if ipnet.Contains(ip1) {
+			return cidrString(ipmask, l)
```
nickgarlis: It's probably easier to create the CIDR without converting it into a string by doing something like this instead.
```
&net.IPNet{
	IP:   ip,
	Mask: mask,
}
```

util.go
tacerus: From my understanding it will just return the address itself, for example calling `Prev()` on `0.0.0.0`, will yield `0.0.0.0`.

util.go
tacerus: Think that would make sense and align with the return being net.IPNet.

util.go
tacerus: Good idea, I think I can even just return the existing `ipnet`.

util.go
nickgarlis: This is the `IntervalOpen` scenario. Consider a CIDR like `128.0.0.0/1`. In that instance, the user would have to call `NetIntervalRange` by setting `last` into an empty IP or nil. 

I'd suggest an empty IP instead of a nil value for consistency with `NetInterval` but I have no strong opinions about this.

util_test.go
```diff
@@ -201,3 +201,135 @@ func TestNetInterval(t *testing.T) {
...
+		name    string
+		first   string
+		last    string
+		wantNet string
+		wantErr bool
+	}{
+		{
+			first:   "0.0.0.1",
+			last:    "255.255.255.254",
+			wantNet: "0.0.0.0/0",
```
nickgarlis: Shoudn't `0.0.0.0/0` be (0.0.0.0 - 255.255.255.255) ?

util.go
tacerus: I see - should I just pass both input values to `NetFromRange` as is if `last` is empty?

util_test.go
tacerus: Good point, and the test passes with 0.0.0.0 - 255.255.255.255 => 0.0.0.0/0 as well. But I guess it is right that 0.0.0.1 - 255.255.255.254 yields 0.0.0.0/0 too?

util.go
nickgarlis: I am not sure. `NetFromRange` is a bit different since it never expects a null or otherwise invalid IP, so returning an error there is reasonable.

`NetFromInterval`, however, is a different situation. An empty last value doesn’t inherently mean the interval is invalid, so treating it as an error wouldn’t be correct. That’s why I’m leaning toward allowing an empty IP to represent the open end of the interval.

util.go
```diff
@@ -126,3 +128,48 @@ func NetInterval(cidr string) (net.IP, net.IP, error) {
...
+			return &ipnet, nil
+		}
+	}
+
+	return nil, errors.New("Failed to construct network from range.")
+}
+
+// NetFromNetInterval returns a CIDR IP network given a start and end address as found in intervals.
+// This is similar to NetFromRange, but subtracts one address from the end of the range.
+func NetFromIntervalRange(first net.IP, last net.IP) (out *net.IPNet, err error) {
```
nickgarlis: nit: perhaps `NetFromInterval` would be a more consistent name with `NetInterval`

util.go
nickgarlis: So what I would do here is, if `last` is null or empty, set `previous` to max IP. Otherwise, calculate the previous IP like you're doing. Perhaps it'd be good to have a test case.

util_test.go
nickgarlis: My understanding is that 0.0.0.1 - 255.255.255.254 can only fit in 0.0.0.0/0 but this shouldn't mean that 0.0.0.0/0 is the exact CIDR for this range. Perhaps your implementation tries to find the best CIDR for the range instead ?

I checked to see how nft chooses to represent this range and it seems to print it as is and not as a CIDR.

You can test by adding these set elements 
```golang
c.AddSet(set, []nftables.SetElement{
	{Key: net.ParseIP("0.0.0.1").To4()},
	{Key: net.ParseIP("255.255.255.255").To4(), IntervalEnd: true},
})
```
which produces
```
set my_set {
	type ipv4_addr
	flags interval
	elements = { 0.0.0.1-255.255.255.254 }
}
```

tacerus: Good catch, I did not consider this. Should we process ranges which do not match an exact CIDR here at all? I think it's not an error if it does not match, but I understand the "nearest" match is not quite expected either - what do you think of returning nil for these?

(Then some later additional logic can decide to print first and last as is, depending on the use case.)

util.go
tacerus: Makes sense, will adjust.

util_test.go
nickgarlis: >  Should we process ranges which do not match an exact CIDR here at all?

I think that those ranges should not return a CIDR.

An option could be to return an `ok` variable. Something like this.
```golang
cidr, ok := nftables.NetFromInterval(first, last)
```
An error would also be fine I guess but then it'd have to be typed so that it can be handled.

util.go
tacerus: I implemented this now, however I unfortunately don't understand why it should be the maximum address.

Edit: if the added logic is correct, what would be an IPv6 example to add to the tests?

util_test.go
tacerus: Added this now and expanded the tests.

I still have it return the nearest network though, do you prefer to have it not return any network if `ok` is `false` anyways?

Another point, for the end address comparison I copied the logic from the `EndIP()` function of another module (https://github.com/3th1nk/cidr/blob/v0.3.0/cidr.go#L154), is that an issue license wise? Unfortunately I'm not sure how to rewrite it.

util.go
nickgarlis: If you have an interval with no explicit last element and the `IntervalOpen` flag is set, you’d end up passing something like (first: 128.0.0.0, last: 0.0.0.0). In this case, you want to interpret it as the full range rather than a half-open one, i.e. (first: 128.0.0.0, last: 255.255.255.255), which can be represented exactly as the CIDR 128.0.0.0/1.

For IPv6, the upper half range would be `8000::/1` (`8000::`, `ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff`).

util_test.go
nickgarlis: > I still have it return the nearest network though, do you prefer to have it not return any network if ok is false anyways?

Yes, I think i'd make more sense for it to be the exact network or nothing.

When it comes to the linked library, it seems to be MIT so it _should_ be fine but I'd ask @stapelberg.

util.go
tacerus: Thanks for the explanation!

util_test.go
tacerus: OK, I adjusted it to return nil if no exact match is found.

nickgarlis (approved):

PR #352, opened 2026-02-10
Use poll instead of pselect6

3u13r:
pselect6 is limited to FD_SETSIZE, which is 1024 in most cases. When a application holds many fds, this can be reached easily, resulting in a panic when the fd is added to the fd set. Instead of pselect6 use poll, which doesn't have such a limitation.

### DIFF
--- a/socket_test.go
+++ b/socket_test.go
@@ -1,0 +1,58 @@
+package nftables_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/google/nftables"
+	"github.com/google/nftables/internal/nftest"
+	"golang.org/x/sys/unix"
+)
+
+// TestIsReadReadyHighFD verifies that the nftables library works correctly when
+// the underlying netlink socket gets an fd >= FD_SETSIZE (1024). The old
+// pselect-based implementation would panic in this scenario because
+// unix.FdSet.Set panics for fd >= FD_SETSIZE. The current poll-based
+// implementation has no such limit.
+func TestIsReadReadyHighFD(t *testing.T) {
+	c, newNS := nftest.OpenSystemConn(t, *enableSysTests)
+	defer nftest.CleanupSystemConn(t, newNS)
+
+	// Exhaust low file descriptors so the next socket allocation gets fd >= FD_SETSIZE.
+	var fillers []*os.File
+	defer func() {
+		for _, f := range fillers {
+			f.Close()
+		}
+	}()
+
+	for {
+		f, err := os.Open("/dev/null")
+		if err != nil {
+			t.Fatalf("os.Open(/dev/null): %v", err)
+		}
+		fillers = append(fillers, f)
+		if int(f.Fd()) >= unix.FD_SETSIZE {
+			break
+		}
+	}
+	t.Logf("exhausted fds up to %d", fillers[len(fillers)-1].Fd())
+
+	// By default, a transient socket is created for each request. The socket
+	// will get an fd >= FD_SETSIZE. With the old pselect code this would panic;
+	// with poll it must work.
+
+	// Add a command and flush it to trigger the isReadReady code path.
+	c.AddTable(&nftables.Table{Name: "test_high_fd", Family: nftables.TableFamilyIPv4})
+	func() {
+		// turn the potential panic into a test failure.
+		defer func() {
+			if r := recover(); r != nil {
+				t.Fatalf("isReadReady panicked for fd >= %d: %v", unix.FD_SETSIZE, r)
+			}
+		}()
+		if err := c.Flush(); err != nil {
+			t.Fatalf("Flush() failed: %v", err)
+		}
+	}()
+}
--- a/socket.go
+++ b/socket.go
@@ -8,7 +8,7 @@ import (
 )
 
 // isReadReady reports whether the netlink connection is ready for reading.
-// It uses pselect6 with a zero timeout on the underlying raw connection.
+// It uses poll(2) with a zero timeout on the underlying raw connection.
 // This allows for an efficient check of socket readiness without blocking.
 // If the Conn was created with a TestDial function, it assumes readiness.
 func (cc *Conn) isReadReady(conn *netlink.Conn) (bool, error) {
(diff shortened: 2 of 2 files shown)


nickgarlis:
@3u13r This makes sense. I wasn’t aware of the FD_SETSIZE limit here. The original idea was to mirror what nft does, but the use cases are a bit different. nft likely never ends up with a large number of open FDs, whereas with this library the parent process could own quite a few more.

I assume you didn’t use ppoll because the signal mask isn’t being used and the timeout is immediate?

3u13r:
> whereas with this library the parent process could own quite a few more.

Yeah, we run into this because we have a lot of tests, each in their own network namespace testing different nftables things. Also those run concurrently.

> I assume you didn’t use ppoll because the signal mask isn’t being used and the timeout is immediate?

Now that you're mentioning it, `ppoll` would probably be the more exact translation. Though I didn't look that deep into it since, as you said, we return immediately anyway.
Do you think this is subject to change? Do you prefer ppoll?

nickgarlis:
> Now that you're mentioning it, ppoll would probably be the more exact translation. Though I didn't look that deep into it since, as you said, we return immediately anyway.
> Do you think this is subject to change? Do you prefer ppoll?

I don’t have a strong preference here. I’m not a maintainer, but using poll seems fine as long as we’re confident there’s no difference compared to ppoll with these inputs. I also don’t see any reason to use a signal mask in this case.

nickgarlis (approved):

stapelberg (approved):

socket_test.go
```diff
@@ -0,0 +1,81 @@
...
+// TestIsReadReadyHighFD verifies that the nftables library works correctly when
+// the underlying netlink socket gets an fd >= FD_SETSIZE (1024). The old
+// pselect-based implementation would panic in this scenario because
+// unix.FdSet.Set panics for fd >= FD_SETSIZE. The current poll-based
+// implementation has no such limit.
+func TestIsReadReadyHighFD(t *testing.T) {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ns, err := netns.New()
```
stapelberg: This test fails on GitHub Actions:

> --- FAIL: TestIsReadReadyHighFD (0.00s)
>    socket_test.go:23: netns.New() failed: operation not permitted
> FAIL

socket_test.go
```diff
@@ -0,0 +1,81 @@
...
+			t.Fatalf("os.Open(/dev/null): %v", err)
+		}
+		fillers = append(fillers, f)
+		if int(f.Fd()) >= unix.FD_SETSIZE {
+			break
+		}
+	}
+	t.Logf("exhausted fds up to %d", fillers[len(fillers)-1].Fd())
+
+	// Confirm FdSet.Set would panic at this fd range (regression anchor).
```
stapelberg: I don’t think we need this check. Can you remove line 47 to 58? The rest of test case is sufficient to ensure many fds work correctly.

stapelberg (changes requested):

3u13r:
Thanks for the fast review(s). I've guarded the test with the `run_system_tests` flag. Not sure if you prefer those the new test as part of the integration tests or in the `run_system_tests` guarded normal tests. Those the second option for now.


Sadly I get testing errors on the main branch when executing `sudo ./nftables.test -test.v -run_system_tests` locally
```
nftables_test.go:8080: conn.Flush() failed: receive: NFT_MSG_NEWTABLE: netlink receive: operation not supported
--- FAIL: TestTableOwnership/owned_persistent_table (0.01s)
(...)
nftables_test.go:8200: expected operation not permitted error, got receive: NFT_MSG_NEWTABLE: netlink receive: operation not supported
--- FAIL: TestSocketDrainingOnErrors/non_short_circuited_error (0.01s)
```
I guess it's due to my kernel version (6.8.0).
But those are the only failures in the integration and normal tests both on main and this branch.

nickgarlis:
> Sadly I get testing errors on the main branch when executing sudo ./nftables.test -test.v -run_system_tests locally
> I guess it's due to my kernel version (6.8.0).

When it comes to the test failures that you're seeing locally, I think you're right to suspect your kernel version as both of these tests use `NFT_TABLE_F_PERSIST` [which is not present in that version]( https://elixir.bootlin.com/linux/v6.8/source/include/uapi/linux/netfilter/nf_tables.h#L183).

socket_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"golang.org/x/sys/unix"
+)
+
+// TestIsReadReadyHighFD verifies that the nftables library works correctly when
+// the underlying netlink socket gets an fd >= FD_SETSIZE (1024). The old
+// pselect-based implementation would panic in this scenario because
+// unix.FdSet.Set panics for fd >= FD_SETSIZE. The current poll-based
+// implementation has no such limit.
+func TestIsReadReadyHighFD(t *testing.T) {
+	if !*enableSysTests {
```
nickgarlis: Perhaps you could use this helper ? https://github.com/google/nftables/blob/1db35da82052e94bf7bb66dd6a2ec05b702a44f0/nftables_test.go#L8496

3u13r: The current order of operation is:
1. Check if we are allowed to run this test
2. Allocate a lot of fds
3. Allocate the netlink fd and some netfilter operations

I can use the helper if you prefer that we move the check in 1. to after the allocation of the 1024 fds.
Of course, we would also allocate the 1024 fds when running the normal tests.

Both options look reasonable to me, do you prefer using the helper?

nickgarlis: So the connection needs to open only after the fds have been allocated but this helper does everything in one step ?

Perhaps you could do something like the following ? 
```golang
_, newNS := nftest.OpenSystemConn(t, *enableSysTests)

// Allocate FDs

c, err := nftables.New(nftables.WithNetNSFd(int(newNS)))
```

Also, reusing the connection created by the helper might still fail when pselect6 is used, since the default behavior is that the socket isn’t reused across operations.

3u13r: Ah didn't see that the socket is not reused. That makes things easier. 
Going with the default test setup that uses `nftest.OpenSystemConn()` then.

nickgarlis (approved):

PR #355, opened 2026-02-11
set: fix byte order metadata for host-endian types and map data

anthonyrisinger:
Fix two issues with set/map userdata byte order metadata that cause `nft list ruleset` to misinterpret element keys and data values on little-endian systems. The kernel-level data is always correct — these are display/round-trip issues affecting the userdata TLVs that the `nft` CLI uses to reconstruct human-readable output.

## Bug 1: Anonymous sets ignore explicit `KeyByteOrder`

Anonymous/constant/interval sets unconditionally emit `KEYBYTEORDER=2` (big-endian), ignoring the `KeyByteOrder` field. The `nft` C tool always uses the key expression's actual byte order ([`mnl.c:mnl_nft_set_add`](https://git.netfilter.org/nftables/tree/src/mnl.c?id=187c6d01d35722618c2711bbc49262c286472c8f#n1165)), not a blanket default.

For host-endian types like `mark_type` (`BYTEORDER_HOST_ENDIAN`) and `ifname_type` (`BYTEORDER_HOST_ENDIAN`), this causes `nft list ruleset` to misinterpret the element data on LE systems:

- **Mark sets**: `meta mark { 0x00000000, 0x01000000 }` instead of `{ 0x00000000, 0x00000001 }`
- **Interface name vmaps**: `iifname vmap { "" : jump chain1 }` instead of `{ "eth0" : jump chain1 }`

The existing comment "Semantically useless - kept for binary compatability with nft" is incorrect — `KEYBYTEORDER` IS semantically meaningful for host-endian types. The `nft` C tool uses it in [`netlink_delinearize_setelem`](https://git.netfilter.org/nftables/tree/src/netlink.c) to decide whether to call `mpz_switch_byteorder` on element keys.

**Fix**: When the anonymous/constant/interval condition fires, check if `KeyByteOrder` is explicitly set to `NativeEndian` and emit `1` (host) instead of `2` (big). Falls back to the original big-endian default when `KeyByteOrder` is unset (zero value), preserving backwards compatibility.

## Bug 2: Maps never emit `DATABYTEORDER`

The `NFTNL_UDATA_SET_DATABYTEORDER` constant exists in `userdata.go` but is never used. The `Set` struct has no `DataByteOrder` field, and the userdata construction has no code path that emits this TLV.

The `nft` C tool [emits `DATABYTEORDER`](https://git.netfilter.org/nftables/tree/src/mnl.c?id=187c6d01d35722618c2711bbc49262c286472c8f#n1169) for all datamaps:

```c
if (set_is_datamap(set->flags))
    nftnl_udata_put_u32(udbuf, NFTNL_UDATA_SET_DATABYTEORDER,
                         set->data->byteorder);
```

Without it, `nft list ruleset` uses `BYTEORDER_INVALID` for map data, skips the `mpz_switch_byteorder` call, and displays native-endian values as byte-swapped on LE systems:

```
# Expected:
elements = { ... : 0x00000001 }

# Without DATABYTEORDER on LE:
elements = { ... : 0x01000000 }
```

**Fix**: Add `DataByteOrder` field to `Set` struct and emit `NFTNL_UDATA_SET_DATABYTEORDER` in the userdata construction for maps when set.

## Discovery context

Found while debugging a WireGuard mesh overlay that uses `meta mark` sets and maps for zone-based connection latching. The incorrect display made it extremely difficult to diagnose whether mark values were being set correctly, since `nft list ruleset` showed byte-swapped values but the kernel data was actually correct (verified via `nft --debug=netlink list ruleset`).

## Testing

Verified on linux/arm64 (little-endian) that after this patch:
- Anonymous `TypeMark` sets display correct values (`0x00000001` not `0x01000000`)
- Map data with `TypeMark` displays correct values
- `nft --debug=netlink list ruleset` shows identical kernel-level bytes before and after (no functional change)
- `GOOS=linux go build .` and `GOOS=linux go vet .` pass cleanly

### DIFF
--- a/nftables_test.go
+++ b/nftables_test.go
@@ -3742,6 +3742,144 @@ func TestCreateAutoMergeSet(t *testing.T) {
 	}
 }
 
+func TestSetByteOrderRoundTrip(t *testing.T) {
+	tests := []struct {
+		name          string
+		set           nftables.Set
+		elems         []nftables.SetElement
+		wantKeyOrder  binaryutil.ByteOrder
+		wantDataOrder binaryutil.ByteOrder
+	}{
+		{
+			name: "constant set defaults key byteorder to big-endian",
+			set: nftables.Set{
+				Constant: true,
+				KeyType:  nftables.TypeInetService,
+			},
+			elems: []nftables.SetElement{
+				{Key: binaryutil.BigEndian.PutUint16(80)},
+			},
+			wantKeyOrder: binaryutil.BigEndian,
+		},
+		{
+			name: "explicit host-endian key byteorder",
+			set: nftables.Set{
+				KeyType:      nftables.TypeMark,
+				KeyByteOrder: binaryutil.NativeEndian,
+			},
+			elems: []nftables.SetElement{
+				{Key: binaryutil.NativeEndian.PutUint32(1)},
+			},
+			wantKeyOrder: binaryutil.NativeEndian,
+		},
+		{
+			name: "interval set defaults key byteorder to big-endian",
+			set: nftables.Set{
+				Interval: true,
+				KeyType:  nftables.TypeInetService,
+			},
+			wantKeyOrder: binaryutil.BigEndian,
+		},
+		{
+			name: "interval set with explicit host-endian key byteorder",
+			set: nftables.Set{
+				Interval:     true,
+				KeyType:      nftables.TypeMark,
+				KeyByteOrder: binaryutil.NativeEndian,
+			},
+			elems: []nftables.SetElement{
+				{Key: binaryutil.NativeEndian.PutUint32(7)},
+			},
+			wantKeyOrder: binaryutil.NativeEndian,
+		},
+		{
+			name: "map with explicit host-endian key and data byteorder",
+			set: nftables.Set{
+				KeyType:       nftables.TypeMark,
+				DataType:      nftables.TypeMark,
+				KeyByteOrder:  binaryutil.NativeEndian,
+				DataByteOrder: binaryutil.NativeEndian,
+				IsMap:         true,
+			},
+			elems: []nftables.SetElement{
+				{
+					Key: binaryutil.NativeEndian.PutUint32(1),
+					Val: binaryutil.NativeEndian.PutUint32(2),
+				},
+			},
+			wantKeyOrder:  binaryutil.NativeEndian,
+			wantDataOrder: binaryutil.NativeEndian,
+		},
+		{
+			name: "map with explicit data byteorder only",
+			set: nftables.Set{
+				KeyType:       nftables.TypeInetService,
+				DataType:      nftables.TypeMark,
+				DataByteOrder: binaryutil.NativeEndian,
+				IsMap:         true,
+			},
+			elems: []nftables.SetElement{
+				{
+					Key: binaryutil.BigEndian.PutUint16(22),
+					Val: binaryutil.NativeEndian.PutUint32(1),
+				},
+			},
+			wantDataOrder: binaryutil.NativeEndian,
+		},
+	}
+
+	for i, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			conn, newNS := nftest.OpenSystemConn(t, *enableSysTests)
+			defer nftest.CleanupSystemConn(t, newNS)
+			defer conn.FlushRuleset()
+
+			table := conn.AddTable(&nftables.Table{
+				Name:   fmt.Sprintf("byteorder-table-%d", i),
+				Family: nftables.TableFamilyIPv4,
+			})
+
+			set := tt.set
+			set.Table = table
+			set.Name = fmt.Sprintf("byteorder-set-%d", i)
+
+			if err := conn.AddSet(&set, tt.elems); err != nil {
+				t.Fatalf("failed to add set: %v", err)
+			}
+			if err := conn.Flush(); err != nil {
+				t.Fatalf("failed to flush: %v", err)
+			}
+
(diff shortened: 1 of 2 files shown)


anthonyrisinger:
Company decided they wanted the PR from here—apologies for the noise from the last two closed!—I couldn't figure out how to open #353 again; it was the same exact commit on all three.

set.go
```diff
@@ -716,12 +717,27 @@ func (cc *Conn) AddSet(s *Set, vals []SetElement) error {
...
 		// Per https://git.netfilter.org/nftables/tree/src/mnl.c?id=187c6d01d35722618c2711bbc49262c286472c8f#n1165
 		userData = userdata.AppendUint32(userData, userdata.NFTNL_UDATA_SET_KEYBYTEORDER, 1)
+	} else if s.Anonymous || s.Constant || s.Interval || s.KeyByteOrder == binaryutil.BigEndian {
+		userData = userdata.AppendUint32(userData, userdata.NFTNL_UDATA_SET_KEYBYTEORDER, 2)
+	}
+
+	// Emit DATABYTEORDER for maps, matching nft C tool behavior (mnl.c:mnl_nft_set_add).
+	// Without this, nft list ruleset cannot determine the data byte order and displays
+	// host-endian values (like marks) as byte-swapped on LE systems.
+	if s.IsMap {
```
nickgarlis, line 734: Shouldn't deserialization also be implemented for `NFTNL_UDATA_SET_DATABYTEORDER` ? This is where it's done is for set elements 
https://github.com/google/nftables/blob/1db35da82052e94bf7bb66dd6a2ec05b702a44f0/set.go#L332

Looks like the implementation for `NFTNL_UDATA_SET_KEYBYTEORDER` is also missing.

nftables_test.go
```diff
@@ -6495,6 +6495,206 @@ func TestVmap(t *testing.T) {
...
+				t.Fatalf("c.AddSet() failed: %v", err)
+			}
+
+			if err := c.Flush(); err != nil {
+				t.Fatalf("c.Flush() failed: %v", err)
+			}
+
+			// Find the NFTA_SET_USERDATA attribute (type 0x0d) in captured messages.
+			// The attribute uses standard netlink TLV: 2-byte length, 2-byte type, then data.
+			const nftaSetUserData = 0x0d
```
nickgarlis: Lines 6652-6685 seem a bit complicated. Perhaps an easier way to do this would be to call `GetSetElements` and check that the values match ? It would require deserialization though (see my other comment).

nickgarlis (commented):
I took a look through this and left some suggestions / personal preferences. @stapelberg I don't mean to step on your toes. Just thought those comments might be helpful.

set.go
anthonyrisinger: Great catch, agreed. I added set-level deserialization in `setsFromMsg` for both `NFTNL_UDATA_SET_KEYBYTEORDER` and `NFTNL_UDATA_SET_DATABYTEORDER`, so map/set metadata now round-trips on read as well as write.

nftables_test.go
anthonyrisinger: Yep, agreed this was too low-level. I replaced that raw `NFTA_SET_USERDATA` byte-scanning test with a simpler round-trip test in `set_test.go` (serialize via `AddSet`, decode via `setsFromMsg`/`elementsFromMsg`, then assert key/data byte-order behavior and values).

anthonyrisinger:
Let me know if there's any other changes—I simplify tests with a round-trip through both encode/decode paths.

set_test.go
```diff
@@ -308,3 +309,184 @@ func TestMarshalSet(t *testing.T) {
 		})
 	}
 }
+
+func TestSetByteOrderRoundTrip(t *testing.T) {
```
nickgarlis: I’m wondering whether this should remain a unit test, as it is now, or be moved to an integration test in nftables_test.go.

Running it without a test dial might give us a clearer picture of how the kernel actually responds to these parameters.

Here’s an example of what an integration test looks like:
https://github.com/google/nftables/blob/main/nftables_test.go#L8495

anthonyrisinger: Are there any parts that should remain here as a unit test, or is it suitable to move in its entirety? I haven't yet looked at the linked integration tests, mostly trying to clarify beforehand, and confirm I'm reading you correctly; I think you're saying it's OK to move all of them and update the pattern accordingly, but if there's something that should be left behind, let me know ☺️

nickgarlis: I think integration tests alone are sufficient :+1:

anthonyrisinger:
Should be all set! Validated with a Docker container:

```console
$ docker run --rm --privileged -v ~/devel/nftables:/work -w /work golang:1.24-bookworm bash -lc '/usr/local/go/bin/go test -count=5 -run "^TestSetByteOrderRoundTrip$" -run_system_tests .'

go: downloading github.com/mdlayher/netlink v1.8.1-0.20251028132421-dcc6cab9a6eb
go: downloading golang.org/x/sys v0.35.0
go: downloading github.com/vishvananda/netns v0.0.4
go: downloading golang.org/x/net v0.43.0
go: downloading github.com/mdlayher/socket v0.5.1
go: downloading golang.org/x/sync v0.6.0
PASS
ok  	github.com/google/nftables	0.022s
```

set_test.go
anthonyrisinger: Should be all set now

nickgarlis (approved):
LGTM

anthonyrisinger:
No real rush on my end (I'm using a private fork at the moment), but is it reasonable to believe this will find its way into the next release, whenever that may be? Mostly looking to check any boxes before I file this away to the back of my mind ☺️

PR #358, opened 2026-02-28
Update dependencies and Go version to 1.25

nickgarlis:
Includes updates to several libraries, including github.com/mdlayher/netlink,
which has some relevant improvements (mainly performance):
- https://github.com/mdlayher/netlink/blob/main/CHANGELOG.md#v1111
- https://github.com/mdlayher/netlink/blob/main/CHANGELOG.md#v1110
- https://github.com/mdlayher/netlink/blob/main/CHANGELOG.md#v1100


go.mod
```diff
@@ -1,17 +1,17 @@
 module github.com/google/nftables
 
-go 1.23.0
+go 1.24.0
```
aojea: is this needed? IIRC this forces the consumers to bump too and for libraries is less disruptive to keep it at the minimum possible

nickgarlis: It's not urgent. It was only needed to use the latest tagged version of mdlayher/netlink because I bumped the versions there.

This is mainly an upkeep task so IMO, it doesn't need to be merged before the next release.

aojea: ah, if a dependency was bumped then it makes sense, sorry about that, is just sometime I found people bumping then without noticing this has the consequence that downstream users need to bump too

PR #362, opened 2026-07-22
ci: add apidiff workflow to detect breaking changes

aojea:
Adds a GitHub Actions job to compare the module API of a PR branch against the target branch. The job uses `apidiff` in module mode to detect incompatible changes or removed packages. Intentional breaking changes can be allowed by adding the 'breaking change allowed' label to the PR (this label needs to be created)

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -35,3 +35,42 @@ jobs:
         sudo ./nftables.test -test.v -run_system_tests
         go test -c github.com/google/nftables/integration
         (cd integration && sudo ../integration.test -test.v -run_system_tests)
+
+  apidiff:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/push.yml
```diff
@@ -35,3 +35,42 @@ jobs:
         sudo ./nftables.test -test.v -run_system_tests
         go test -c github.com/google/nftables/integration
         (cd integration && sudo ../integration.test -test.v -run_system_tests)
+
+  apidiff:
+    runs-on: ubuntu-latest
+    # Run only on pull requests, and skip if the PR has the 'breaking change allowed' label.
+    if: ${{ github.base_ref && !contains(github.event.pull_request.labels.*.name, 'breaking change allowed') }}
```
aojea, line 42: the label needs to be created to avoid having to override the job

